PR #309, opened 2014-09-15
Make clearer in README how sudoers file should be modified.

nhstanley:
Adding the suggested line in the wrong place will cause you to
be locked out after initial deployment.

Arguably you should know what you're doing with these things,
but for those who don't have deep background in sys admin this
can be an unexpected gotcha (Ansible pre 1.7 does not error).


al3x:
Would like to merge a PR that uses the option @malvim suggested.

nhstanley:
Closing because it should be solved by #343.

PR #319, opened 2014-11-01
Update roundcube role to use wheezy backports

mikeashley:
Debian 7.7 uses Roundcube 0.7.x, which is pretty old.  There appear to be lots of fixes and improvements in the 0.9.x and 1.0.x series.  Suggest moving to 0.9.x using the package in wheezy backports.

Changes:
- Added backports to apt configuration.
- Updated roundcube task to use the roundcube backport.
- Updated roundcube's main.inc.php configuration file to use the one included in the backport distribution.  Migrated configuration parameters from the existing configuration file except in cases where it looked like the Roundcube team was changing defaults that were not touched by Sovereign's configuration.  I may have gotten some wrong; needs review by a Sovereign maintainer.

This was tested in the development environment and has been deployed on my own server.

### DIFF
--- a/roles/webmail/files/etc_roundcube_main.inc.php
+++ b/roles/webmail/files/etc_roundcube_main.inc.php
@@ -6,7 +6,10 @@
 |                                                                       |
 | This file is part of the Roundcube Webmail client                     |
 | Copyright (C) 2005-2011, The Roundcube Dev Team                       |
-| Licensed under the GNU GPL                                            |
+|                                                                       |
+| Licensed under the GNU General Public License version 3 or            |
+| any later version with exceptions for skins & plugins.                |
+| See the README file for a full license statement.                     |
 |                                                                       |
 +-----------------------------------------------------------------------+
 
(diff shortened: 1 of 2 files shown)


iboxifoo:
I'm having issues on an existing install on Debian when running the playbook. I'm pretty sure the task is hanging here (i.e. I left it for almost an hour):

> TASK: [webmail | Install Roundcube from wheezy-backports]

When I first ran the playbook I received an email from debconf. Subject line was:

> Configuring roundcube-core -- Empty passwords unsupported with PostgreSQL

Don't know if these were connected because later I received a cron warning there was signing issues with wheezy backports.

Wonder if anyone can give some pointers here. Thanks.

kigero:
I ended up having the same problem this morning; here's how I fixed it.

1. Killed the apt-get process.  In hindsight this was a bad move (because it left the dpkg processes running), so try skipping this step.
1. Doing a *ps|grep dpkg* showed three relevant processes:
  1. /usr/bin/dpkg --force-confdef --force-confold --status-fd 21 --configure roundcube-pgsql:all roundcube-core:all roundcube-plugins:all roundcube:all
  1. /usr/bin/perl -w /usr/share/debconf/frontend /var/lib/dpkg/info/roundcube-core.postinst configure 0.7.2-9+deb7u1
  1. /bin/sh /var/lib/dpkg/info/roundcube-core.postinst configure 0.7.2-9+deb7u1
1. I killed the third one, the roundcube postinit script.  This cleanly stopped the perl and dpkg processes (and probably would have allowed for a graceful exit of the apt-get process, but I don't know what ansible would have done).
1. I then ran *dpkg --configure -a* to try completing the roundcube install.
  1. I refused the dbconfig-common installation, as I figured the db was already set up.
  1. I kept the local configuration file - the only differences between the local file and the new file were that the new file didn't have the db information.
  1. It asked to do a db update, I selected yes.
  1. It then asked for the password of the db admin user, with a note that this should not be required for postgres.  However, entering an empty password displays the error "Empty passwords unsupported with PostgreSQL" (this was most likely the cause of the hang).  So instead I entered the postgres admin password.
  1. Finally I allowed the update of main.inc.php.

After this process completed, roundcube appeared to be updated and working, so I ran the ansible playbook again and this time it completed without error.

ventolin:
Thanks for the detailed writeup, @kigero! I ran into the same issue and was able to solve it by following your example. I skipped step 1 as you recommended and the apt-get process did stop on its own (ansible exited with errors).

PR #336, opened 2014-12-14
fail2ban support for Trusty

mariusv:
This is addressing the issue https://github.com/al3x/sovereign/issues/331 also I had the same issue and tested in  different Ubuntu Trusty environment. Issue causes fail2ban not being able to start.


neuhaus:
Nice, thanks.

PR #339, opened 2015-01-10
More secure defaults for ssh.

fengor:
Hardened SSH configuration as per https://stribika.github.io/2015/01/04/secure-secure-shell.html

Ciphers, Kex and MAC can be set via defaults.var


lukecyca:
Looks good to me. Can we get one other person to review this and/or try it out before I merge?

fengor:
Added some tasks to set Timezone to UTC.

al3x:
@fengor It looks like there was some debate about the recommendations in that post which resulted in some changes. Would you mind incorporating the latest changes (ex: including GCM)?

fengor:
couldnt get the @openssh.com ciphers to work with the default openssh packaged with my debian on the VPS so i trimmed the cipherlist to those included with it.

roles/common/templates/ssh.j2
```diff
@@ -0,0 +1,10 @@
+    Ciphers {{ ciphers }}
+    #KexAlgorithms curve25519-sha256@libssh.org,diffie-hellman-group-exchange-sha256
```
al3x: Please remove commented line.

roles/common/tasks/security.yml
```diff
@@ -16,6 +16,10 @@
 - name: Ensure fail2ban is started
   service: name=fail2ban state=started
 
-- name: Update sshd config to disallow root logins
-  lineinfile: dest=/etc/ssh/sshd_config regexp=^PermitRootLogin line="PermitRootLogin no" state=present
+- name: Update sshd config for PFS and more secure defaults
+  template: src=sshd.j2 dest=/etc/ssh/sshd_config
```
al3x: This line should follow our conventions for file names, which include the parent directories:

    template: src=etc_ssh_sshd_config.j2 dest=/etc/ssh/sshd_config

roles/common/tasks/security.yml
```diff
@@ -16,6 +16,10 @@
...
   notify: restart ssh
+  
+- name: Update ssh config for more secure defaults
+  template: src=ssh.j2 dest=/etc/ssh/ssh_config
```
al3x: As in the previous task, please follow file name conventions:

     template: src=etc_ssh_ssh_config.j2 dest=/etc/ssh/ssh_config

cweagans:
:+1: LGTM

PR #343, opened 2015-01-15
Automatically set sudoers using /etc/sudoers.d/

nhstanley:
This commit makes it so that sudo privileges are automatically set for the main user account via /etc/sudoers.d/. I have tested this on Debian Wheezy with Vagrant, and it appears to work as expected (running e.g. `"sudo vim /etc/sudoers"` works without password, as does `"sudo su -"`). This should make obsolete pull request  #309 (also from me, will close). I have also removed from the README the lines about adding this functionality via `visudo`, as it is no longer needed.

Please note I am not an expert on this, so it would be good to have someone look this over/test it out and ensure that I haven't missed something. If I did this in a non-ideal way, I'd like to learn how to do it better next time. This is my first real OSS commit via git/github, so accept with caution.

### DIFF
--- a/README.textile
+++ b/README.textile
@@ -123,10 +123,6 @@ nano /home/deploy/.ssh/authorized_keys
 chmod 400 /home/deploy/.ssh/authorized_keys
 chown deploy:deploy /home/deploy -R
 
-This account should be set up for passwordless sudo. Use @visudo@ and add this line:
-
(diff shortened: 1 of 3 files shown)


neuhaus:
Good idea

poops:
Having the leading + in the sudoers file doesn't work for me in Debian 7.8. Using sudo still prompts me for a password. Removing the + solves the issue for me.

PR #345, opened 2015-01-19
modifying the sysctl,fetch and remove some unnecessary tasks

arbabnazar:
- Generate the individual cert/key/csr for each client with their name instead of using client for everything
- Modify the "sysctl"
- We don't need to copy the ca.crt and ta.key files into the clients config directories because we already register them and add them to the ovpn file
- Fetch the single ovpn files for individual client that contains everything and save them into the clients directory in the same place, where we will run the playbook

Verified the downloaded ovpn file(s) on Windows and Linux clients.

Thanks for writing such a wonderful playbook.


arbabnazar:
@al3x, please review the changes, hope this will help. Thanks

- Install the iptables-persistent to save the iptables rule instead of adding it to the rc.local
- Get the result of iptables -L -v command and register the result so that we can add and save the iptables rules if it doesn't have them already
- Instead of copy the rc.local as template, adding the dnsmasq service's restart line in it becasue may be there are another lines in it and accidently the provided template overwrite those lines on the production server
- Change the prompt message

lvillani:
I'm not sure whether `iptables-persistent` plays well with `ufw`.

arbabnazar:
@lvillani thanks for the comment, I don't think so it can create any problem as here it will just save the defined rule. Can you please tell me, what kind of problem it can create if they work along. Thanks

ariddell:
Looks good. Rather than add a dependency on `iptables-persistent`, what about using `ufw`, e.g., writing to `/etc/ufw/before.rules`. See http://ubuntuforums.org/showthread.php?t=1250127

arbabnazar:
@ariddell thanks, definitely look into this and will update it. regards

al3x:
@arbabnazar Would still love to incorporate this PR with `ufw` support! :smile_cat:

arbabnazar:
@al3x, you want me to remove the iptables-persistent or use it with ufw. Please elaborate. Thanks

al3x:
I'd like you to remove iptables-persistent, using ufw in its place.

al3x:
Please reopen if you can contribute a patch using ufw.

PR #347, opened 2015-01-31
Use znc distribution package on Ubuntu Trusty

neuhaus:
I haven't tested this yet, just putting it up here for refinement. I'm making some room for some VMs so I can test it with both Wheezy and Trusty (and perhaps Jessie).


al3x:
We do need to move to using packaged versions of znc. This PR is way out of date, unfortunately, so I'm closing it.

PR #366, opened 2015-04-16
adds deploy user to sudoers

poops:
In order for `ansible-playbook -i ./hosts site.yml` to work, you need to add the deploy user to sudoers, otherwise you need to pass the -K flag to `ansible-playbook`

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ Authorize your ssh key if you want passwordless ssh login (optional):
     nano /home/deploy/.ssh/authorized_keys
     chmod 400 /home/deploy/.ssh/authorized_keys
     chown deploy:deploy /home/deploy -R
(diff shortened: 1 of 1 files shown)


simontretter:
deploy should not be hardcoded as some people might  change it in site.yml to another username.

poops:
@simontretter this is added to the documentation section where deploy is already hardcoded, and is optional to setup.

cimm:
For what it's worth: the `/etc/sudoers.d` directory did not exist on my Debian 8.2 box. I had to create it before running the command.

neuhaus:
Isn't sudo installed by default? If not, @cimm please submit a pull request :+1:

cimm:
@neuhaus Was testing it on Debian 8, not 7 as it says in the README, my bad. Sorry.

PR #386, opened 2015-06-13
Remove duplication with items unattended upgrades

spk:
Already present [here](https://github.com/sovereign/sovereign/blob/master/roles/common/tasks/main.yml#L31)


neuhaus:
Thanks.

spk:
:beers:

Issue #341, opened 2015-01-15
RC4 considered broken


neuhaus:
RC4 has had some bad press lately, I think we should disable it (in Apache etc.). This breaks only IE6 on Windows XP afaik.
If there are no objections I'll make a PR.

stephenjamieson:
it could break on some old androids, but in my opinion that's ok.

neuhaus:
Android 2.3.7 doesn't even support [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) which means we are already incompatible.

I have a site configured without RC4 and according to https://www.ssllabs.com/ssltest/ the only `Protocol or cipher suite mismatch` is with IE6+XP.

al3x:
@neuhaus Please do submit a PR :sunglasses:

neuhaus:
Not sure what the best `SSLCipherSuite` is (there doesn't seem to be a big difference - i have these three in use they all do the trick:

    SSLCipherSuite ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AES:RSA+3DES:!ADH:!AECDH:!MD5:!DSS
    SSLCipherSuite "EECDH+AESGCM EDH+AESGCM EECDH -RC4 EDH -SEED !aNULL !eNULL !LOW !3DES !AES128 !MD5 !EXP !PSK !SRP !DSS !RC4 HIGH"
    SSLCipherSuite "EECDH+AESGCM EDH+AESGCM EECDH -RC4 EDH -SEED !aNULL !eNULL !LOW !3DES !MD5 !EXP !PSK !SRP !DSS !RC4 HIGH"

neuhaus:
Mozilla has some recommendations and a config builder at https://wiki.mozilla.org/Security/Server_Side_TLS

Here is the recommended SSLCipherSuite for the "modern" setting:
[...]

PR #389, opened 2015-06-16
Use "modern" SSLCipherSuite per Mozilla recommendations.

neuhaus:
See https://wiki.mozilla.org/Security/Server_Side_TLS for details.
Removes RC4 cipher. Fixes issue #341.
Also explicitly disabled SSLCompression and enables OCSP stapling.

We should put all these settings in
/etc/apache2/mods-enabled/ssl.conf
to avoid duplication...


neuhaus:
OK, need to remove the stapling options then. I can't do it before Tuesday.

PR #372, opened 2015-05-21
Dovecot: Fix for logjam attack

neuhaus:
See also: https://weakdh.org/sysadmin.html

PR #391, opened 2015-06-21
Update dovecot version from wheezy backports

iboxifoo:
For correct implementation of the fix for logjam attack (https://github.com/sovereign/sovereign/pull/372), state=latest is needed to grab sufficient version of Dovecot. If not then https://github.com/neuhaus/sovereign/commit/37aa7e2cb5eb5cf89eb521080c96cd4b8bf4248e doesn't work.


neuhaus:
Nice find, thanks.

PR #393, opened 2015-06-23
Tagged 'deps' aptitude tasks

moul:
Hello,

I'm trying to create a ready-to-use cloud image on Scaleway.
My commit adds a tag 'deps' on all aptitude tasks, so I can easily run the non-dynamic tasks by calling something like: `ansible-playbook --tags=deps site.yml`.
Then, the user can edit `vars` and continue the installation with a lots of time saved.

I don't break the original tagging feature and if you call `ansible-playbook --tags=vpn` it will call all vpn tasks (deps and non-deps ones).

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,9 @@ To run just one or more piece, use tags. I try to tag all my includes for easy i
 
 You might find that it fails at one point or another. This is probably because something needs to be done manually, usually because there’s no good way of automating it. Fortunately, all the tasks are clearly named so you should be able to find out where it stopped. I’ve tried to add comments where manual intervention is necessary.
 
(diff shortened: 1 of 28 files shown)


al3x:
Thanks so much, this is very useful. Before we merge, would you mind adding a quick explanation of how to use the `deps` tag to the `README.md`? Just want to make sure that other people see it and benefit from your work ☺️

moul:
Done. Tell me if it's ok for you.
I should have a real example to show up soon.

ariddell:
What about a slightly more verbose "dependencies" or "depends"?

moul:
:+1: for more verbose, I prefer "dependencies"

PR #396, opened 2015-07-16
Fix apache SSL config for wheezy

mccutchen:
On wheezy, disable the apache SSL config changes related to stapling (introduced #389).

I also took the liberty of consolidating the repeated SSL configuration in a single include file, though I can back that out if that's an unwelcome change.  It's probably worth drawing your attention to the location I chose for this new SSL config:  `/etc/apache2/ssl.conf`

Some other obvious locations I considered:

- `/etc/apache2/conf.d/ssl.conf` — this won't work because anything in this directory apparently gets included by default in `httpd.conf`, and this leads to an invalid apache because of (I think) repeated `SSLEngine on` directives (I can dig up the specific error if that would be useful)
- `/etc/apache2/conf-available/ssl.conf` — on wheezy, the `conf-available` directory is present _after_ a new machine has been provisioned, but it's not present at the point where I'm writing this config file.  It would obviously be easy to add a task to create that directory, if that would be preferable!

This is my first brush w/ sovereign, and I ran into this issue and one more (for which I'll open another PR) while trying to play around with the dev VM.

I hope these changes are useful!


jplock:
LGTM, thanks!

PR #397, opened 2015-07-16
Add missing OpenDMARC dependencies

mccutchen:
When setting up a sovereign dev VM for the first time, I ran into a problem with some missing dependencies and a missing var for the OpenDMARC setup.

These changes allowed the VM to finish provisioning successfully, but I have not yet gotten as far as, you know, verifying that OpenDMARC is working as expected.  Any advice on how best to do that would be appreciated!

### DIFF
--- a/roles/mailserver/tasks/dmarc.yml
+++ b/roles/mailserver/tasks/dmarc.yml
@@ -1,5 +1,11 @@
-- name: Install OpenDMARC milter
-  apt: pkg=opendmarc state=installed update_cache=yes
+- name: Install OpenDMARC milter and related packages
+  apt: pkg={{item}} state=installed
+  with_items:
+    - mysql-server
+    - opendmarc
+    - python-mysqldb
+  tags:
+    - dependencies
 
 - name: Copy OpenDMARC configuration file into place
   template: src=etc_opendmarc.conf.j2 dest=/etc/opendmarc.conf owner=root group=root
@@ -35,4 +41,3 @@
 
 - name: Activate OpenDMARC report cronjob
   cron: name="OpenDMARC report" hour="2" minute="0" job="/bin/bash /etc/opendmarc/report.sh >> /var/log/opendmarc_report.log"
-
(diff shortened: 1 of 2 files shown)


roles/mailserver/tasks/dmarc.yml
```diff
@@ -1,5 +1,11 @@
-- name: Install OpenDMARC milter
-  apt: pkg=opendmarc state=installed update_cache=yes
+- name: Install OpenDMARC milter and related packages
+  apt: pkg={{item}} state=installed
+  with_items:
+    - mysql-server
```
mccutchen: Installing something like mysql might warrant a whole separate task, and might need some ancillary tasks for, say, configuring it securely.  What do y'all think?

roles/mailserver/tasks/dmarc.yml
```diff
@@ -1,5 +1,11 @@
...
+    - opendmarc
+    - python-mysqldb
```
mccutchen: This `python-mysqldb` dependency is actually for the `mysql_db` and `mysql_user` _Ansible_ modules, not for OpenDMARC itself.  I'd also see maybe installing it in a different task (maybe over here in [common/main.yml][1]?), with a comment explaining its use.

[1]: https://github.com/sovereign/sovereign/blob/master/roles/common/tasks/main.yml#L19-L42

jplock:
Will OpenDMARC work with postgres as that's what we're installing by default?

mccutchen:
@jplock I think that's theoretically possible, given its dependency on OpenDBX (from [the docs][1]):

> To compile and operate, this package requires the following:
>
> - OpenDBX (http://www.linuxnetworks.de/doc/index.php/OpenDBX) which acts as a middleware layer between OpenDMARC and your SQL backend of choice

But the schema and OpenDMARC config that's already committed in this repo assume MySQL and I couldn't find any useful info on using Postgres instead.  [This unanswered question on the mailing list][2] from 2013 is the only reference I could find, though maybe someone else around here has some expertise and/or better Google-fu than me!

[1]: http://www.trusteddomain.org/opendmarc/README
[2]: http://www.trusteddomain.org/pipermail/opendmarc-users/2013-October/000194.html

sethetter:
+1

nicbarker:
+1, this branch fixes the OpenDMARC install process for me.

PR #407, opened 2015-08-17
Couple issues with OpenDMARC on Debian 7:

hippich:
Yesterday trying to upgrade Sovereign on my debian 7 VPS along few issues, I got problem install opendmarc. I fixed it with changes in this PR.

 * fix mail_db_opendmarc_username/mail_db_opendmarc_password variable  not found.
 * python-mysqldb package is required. Add it to opendmarc task.

### DIFF
--- a/vars/user.yml
+++ b/vars/user.yml
@@ -88,5 +88,3 @@ selfoss_password_hash: "TODO"
 wallabag_salt: TODO
 wallabag_db_password: TODO
 
-mail_db_opendmarc_username: opendmarc
-mail_db_opendmarc_password: TODO
--- a/vars/user.yml
+++ b/vars/user.yml
@@ -87,4 +87,3 @@ selfoss_password_hash: "TODO"
 # wallabag
 wallabag_salt: TODO
 wallabag_db_password: TODO
-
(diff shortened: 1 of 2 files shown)


vars/user.yml
```diff
@@ -87,3 +87,6 @@ selfoss_password_hash: "TODO"
 # wallabag
 wallabag_salt: TODO
 wallabag_db_password: TODO
+
+mail_db_opendmarc_username: opendmarc
```
miloshadzic: This username is already in ``defaults.yml``. Is it expected that the user will change it?

hippich:
Hm, did not realize it. Somehow after merging latest branch I did not get these in default.yml.. I guess this should be closed then. Thanks for looking into it!

miloshadzic:
@hippich no need to close it IMHO. There are still issues with OpenDMARC and your PR fixes at least one(the missing python-mysqldb package required by ansible).

mccutchen:
See also #397

miloshadzic:
@mccutchen I saw it. I actually tried to rework OpenDMARC for postgres but it's too much work.

PR #422, opened 2015-10-30
Change shell to copy content

spk:
Hi, Just removing `shell` to `copy` with `content` option that IMHO is cleaner.
Cheers

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -42,7 +42,7 @@
     - dependencies
 
 - name: Set timezone to UTC
-  action: shell echo Etc/UTC > /etc/timezone
+  copy: content='Etc/UTC' dest=/etc/timezone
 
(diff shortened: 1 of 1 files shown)


spk:
But no idempotent

nodiscc:
@spk May I ask in what case this would break? Trying to learn more about ansible and your patch looks fine to me.

spk:
Hi @nodiscc, it does not break anything but even if this not changed the `dest` ansible will see as changed every time we play the playbook, it should be idempotent if not change has been made.
Maybe there is a solution on the Internet

spk:
Yes https://github.com/mtpereira/ansible-common/blob/8e693b3db3b15c440428f378e13ce8d04186d637/tasks/timezone-debian.yml

spk:
@nodiscc I've submitted another PR, thanks for the question it help me to fix the problem!

PR #423, opened 2015-10-30
Add some info on DMARC

pjaspers:
I ran a check on http://mail-tester.com and found out I was missing a header to be set for DMARC. SInce I missed it, maybe someone else missed it as well, so maybe it could go into the README.


ariddell:
This looks good to me.

PR #424, opened 2015-11-03
Use common_timezone and fix idempotence

spk:
Thanks-to: https://github.com/mtpereira/ansible-common/commit/8e693b3db3b15c440428f378e13ce8d04186d637

cf https://github.com/sovereign/sovereign/pull/422

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -41,15 +41,22 @@
   tags:
     - dependencies
 
-- name: Set timezone to UTC
-  action: shell echo Etc/UTC > /etc/timezone
+- name: timezone - configure /etc/timezone
+  copy:
+    content: "{{ common_timezone | regex_replace('$', '\n') }}"
+    dest: /etc/timezone
+    owner: root
+    group: root
+    mode: 0644
+  register: common_timezone_config
 
-- name: Set localtime to UTC
+- name: timezone - Set localtime to UTC
   file: src=/usr/share/zoneinfo/Etc/UTC dest=/etc/localtime
+  when: common_timezone_config.changed
 
-- name: Reconfigure tzdata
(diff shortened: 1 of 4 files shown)


roles/common/tasks/main.yml
```diff
@@ -41,15 +41,22 @@
...
+  register: common_timezone_config
 
-- name: Set localtime to UTC
+- name: timezone - Set localtime to UTC
   file: src=/usr/share/zoneinfo/Etc/UTC dest=/etc/localtime
+  when: common_timezone_config.changed
 
-- name: Reconfigure tzdata
-  action: command dpkg-reconfigure -f noninteractive tzdata
-  when: ansible_distribution == 'Debian' or ansible_distribution == 'Ubuntu'
```
spk: Since sovereign is focusing on Debian/Ubuntu I've removed this condition

neuhaus:
Looks good to me but I don't have time to test it

vars/defaults.yml
```diff
@@ -5,6 +5,7 @@
 ###############################################################################
 
 # # common
+# common_timezone: 'Etc/UTC' (required)
```
ariddell: Is there a reason we can't set to UTC as a default?

spk: No don't think so its already currently the default. I will set it as default then.

ariddell: Why mark it (required) then? I think most people would be 

Sorry, I wasn't clear. It's including the line in user.yml that seems
unnecessary -- it implies users need to override it in general. Why not
drop the line from user.yml (i.e., have it be a variable where the
default is fine, as is the case for ``main_user_shell: "/bin/bash"``, which is only
in defaults and not in user.yml)

spk: Yes, see 6330705

ariddell:
This looks good. Thanks!

PR #427, opened 2015-11-05
Z-push apache config: Jessie also uses conf-available/conf-enabled

sread:
Debian Jessie now also uses the a2enconf and conf-available/conf-enabled system, so there is no need to have separate entries for jessie and trusty any longer.

Tested only on jessie but nothing changes for trusty.

### DIFF
--- a/roles/mailserver/tasks/z-push.yml
+++ b/roles/mailserver/tasks/z-push.yml
@@ -41,19 +41,12 @@
 - name: Copy z-push's config.php into place
   template: src=usr_share_z-push_config.php.j2 dest=/usr/share/z-push/config.php
 
-- name: Configure z-push apache alias and php settings
-  copy: src=etc_apache2_conf.d_z-push.conf dest=/etc/apache2/conf.d/z-push.conf
-  notify: restart apache
-  when: ansible_distribution_release != 'trusty'
-
-- name: Create z-push apache alias and php configuration file for Ubuntu Trusty
+- name: Create z-push Apache alias and PHP configuration file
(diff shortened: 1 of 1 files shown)


danmilon:
Yup, :+1: I bumped into this also. Should be ok to merge.

neuhaus:
We're also supporting Wheezy. This change will break Wheezy I think.

sread:
I can definitely do that, but before i do: a lot of the changes in this branch by @al3x from Sept (e.g. 69abd70297a89051c3caa40b304dc134c53d048d) are already incompatible with Wheezy as far as I can tell. So perhaps we can clarify the goals of each branch / long term...

neuhaus:
OK that's news to me (that the jessie branch breaks compatibility with wheezy) but I can live with it. Git branches are so useful :+1: 
It will however make maintaining changes across branches a bit more tedious.

sread:
Yes I was starting worry about the same thing, as some fixes on master are relevant to jessie but not applied and so the two branches are diverging. 

As for this PR, master already special cases wheezy vs trusty. There is no difference in handling between trusty and jessie so I think if we just apply to jessie branch it should be fine (for now).

(That's a lot of code names in one sentence...)

neuhaus:
Just wait for April when we add Xenial to the mix...

neuhaus:
No need to break compatability for this, it's easy to change this in a way that won't break on wheezy:

    ansible_distribution_release == 'trusty' or ansible_distribution_release == 'jessie'

please change the pull request accordingly.

sread:
@neuhaus Done!

neuhaus:
Thanks!

Issue #236, opened 2014-05-06
DNS lookup does not forward through OpenVPN connection


reedlaw:
You can confirm here: https://www.dnsleaktest.com/

I tried changing `/etc/openvpn/server.conf` to use various combinations `bypass-dhcp` and `push "dhcp-option DNS 208.67.222.222"` options, but I was not able to get it working. I also tried setting up a `bind` DNS forwarder, but I'm not sure if `dnsmasq` is already responsible for that.

PR #429, opened 2015-11-14
Improve openvpn security

Friz-zy:
Variables:
- change `chiper` to "AES-256-CBC"
- change `auth` to "SHA512"
- add `verb` level
- add `tun-mtu` for disguising
- add `tls-version-min` and `tls-cipher`

Config:
- add google dns
- add new variables

Src:
- blog.g3rt.nl/openvpn-security-tips.html
- community.openvpn.net/openvpn/ticket/401
- witch.valdikss.org.ru

### DIFF
--- a/roles/vpn/templates/etc_openvpn_server.conf.j2
+++ b/roles/vpn/templates/etc_openvpn_server.conf.j2
@@ -306,7 +306,11 @@ verb {{ openvpn_verb }}
 # category will be output to the log.
 ;mute 20
 
-# Mask your settings with mtu
+# Openvpn changes length of network packages
+# in the certain chunk that depends on
+# chipher and hash-sum algorithms.
+# So, mask your settings by using a lower mtu ;)
+# You can check parameters at this site:
 # witch.valdikss.org.ru
 tun-mtu {{ openvpn_mtu }}
 
--- a/roles/vpn/templates/etc_openvpn_server.conf.j2
+++ b/roles/vpn/templates/etc_openvpn_server.conf.j2
@@ -188,9 +188,9 @@ ifconfig-pool-persist ipp.txt
 # or bridge the TUN/TAP interface to the internet
 # in order for this to work properly).
 ;push "redirect-gateway def1 bypass-dhcp"
+;push "dhcp-option DNS 8.8.8.8"
 push "redirect-gateway def1"
 push "dhcp-option DNS 10.8.0.1"
-push "dhcp-option DNS 8.8.8.8"
 
 # Certain Windows-specific network settings
 # can be pushed to clients, such as DNS
--- a/roles/vpn/templates/etc_openvpn_server.conf.j2
+++ b/roles/vpn/templates/etc_openvpn_server.conf.j2
@@ -190,6 +190,7 @@ ifconfig-pool-persist ipp.txt
 ;push "redirect-gateway def1 bypass-dhcp"
 push "redirect-gateway def1"
 push "dhcp-option DNS 10.8.0.1"
+push "dhcp-option DNS 8.8.8.8"
 
 # Certain Windows-specific network settings
 # can be pushed to clients, such as DNS
@@ -298,9 +299,18 @@ status openvpn-status.log
 # 4 is reasonable for general usage
 # 5 and 6 can help to debug connection problems
 # 9 is extremely verbose
-verb 3
+verb {{ openvpn_verb }}
 
 # Silence repeating messages.  At most 20
 # sequential messages of the same message
 # category will be output to the log.
 ;mute 20
+
+# Mask your settings with mtu
+# witch.valdikss.org.ru
+tun-mtu {{ openvpn_mtu }}
+
+# Set TLS settings
+# Only for openvpn 2.3.3 and >2.3.4
+{{ openvpn_tls_version_min }}
+{{ openvpn_tls_cipher }}
--- a/vars/defaults.yml
+++ b/vars/defaults.yml
@@ -81,6 +81,8 @@ owncloud_db_database: owncloud
 tarsnap_version: 1.0.35
 
 # vpn
+# Notes about security: https://blog.g3rt.nl/openvpn-security-tips.html
+# Check privacy: http://witch.valdikss.org.ru/
 # openvpn_key_country: (required)
 # openvpn_key_province: (required)
 # openvpn_key_city: (required)
@@ -89,8 +91,8 @@ tarsnap_version: 1.0.35
 openvpn_days_valid: "1825"
 openssl_request_subject: "/C={{ openvpn_key_country }}/ST={{ openvpn_key_province }}/L={{ openvpn_key_city }}/O={{ openvpn_key_org }}/OU={{ openvpn_key_ou }}"
 openvpn_key_size: "2048"
-openvpn_cipher: "BF-CBC"
-openvpn_auth_digest: "SHA1"
+openvpn_cipher: "AES-256-CBC"
+openvpn_auth_digest: "SHA512"
 openvpn_path: "/etc/openvpn"
 openvpn_ca: "{{ openvpn_path }}/ca"
 openvpn_dhparam: "{{ openvpn_path }}/dh{{ openvpn_key_size }}.pem"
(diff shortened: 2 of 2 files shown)


vars/defaults.yml
```diff
@@ -98,6 +100,11 @@ openvpn_hmac_firewall: "{{ openvpn_path }}/ta.key"
 openvpn_server: "{{ domain }}"
 openvpn_port: "1194"
 openvpn_protocol: "udp"
+openvpn_mtu: "1300"
+openvpn_verb: "3" # "0" for anonymity
+# uncomment for openvpn 2.3.3 and >2.3.4
+openvpn_tls_version_min: "" # "tls-version-min 1.2"
```
ariddell: Should 1.2 be specified here?

ariddell:
Thanks for this!

(I've been (slowly) working on figuring out the changes needed for jessie in #401)

vars/defaults.yml
Friz-zy: As I learned by my own server, openvpn <= 2.3.2  does not understand TLS options, but it can seriously improve security. So by default this options will be add blank lines into config file.
For newest openvpn it should be:
- openvpn_tls_version_min: "tls-version-min 1.2"
- openvpn_tls_cipher: "tls-cipher TLS-ECDHE-RSA-WITH-AES-128-GCM-SHA256:TLS-ECDHE-ECDSA-WITH-AES-128-GCM-SHA256:TLS-ECDHE-RSA-WITH-AES-256-GCM-SHA384:TLS-DHE-RSA-WITH-AES-256-CBC-SHA256"

Friz-zy:
Thanks for your work ;)
If I have a little more time, I'll try to join because I know ubuntu better than centos.

roles/vpn/templates/etc_openvpn_server.conf.j2
```diff
@@ -190,6 +190,7 @@ ifconfig-pool-persist ipp.txt
 ;push "redirect-gateway def1 bypass-dhcp"
 push "redirect-gateway def1"
 push "dhcp-option DNS 10.8.0.1"
+push "dhcp-option DNS 8.8.8.8"
```
ariddell: If people want to add 8.8.8.8 they typically do it on the VPN client side, right? I personally don't mind adding Google's DNS here but other people might.

vars/defaults.yml
ariddell: Makes sense. Openvpn in Jessie is 2.3.4,FWIW

roles/vpn/templates/etc_openvpn_server.conf.j2
Friz-zy: It is designed for users without experience.
The majority of users even don't know what is DNS ;)

ariddell: The 10.8.0.1 is there to allow the openvpn client to use the server's dnsmasq, right? If 8.8.8.8 needs to be added as a default DNS then it should be done in the dnsmasq configuration, I think. I'm not sure it makes sense to add it here.

I did use the VPN at one point quite a bit and I never had dns problems. This makes me think the server dnsmasq was working fine.

Friz-zy: I use pure openvpn without additional dns server...
I think many installations does not contain dns server. Especially note that your etc_dnsmasq.conf does not contain any necessary options. In my opinion it should be moved to additional role with variables and flexible config template.

ariddell: I think hard coding Google's DNS servers into sovereign might run modestly contrary to the project's spirit. Perhaps the line could be removed?

Separating out dnsmasq might make sense. Is it required for anything besides openvpn?

Friz-zy: I'll leave google dns as comment into config.
Dnsmasq isn't required even for openvpn server.

ngkogkos: Date: Sun, 15 Nov 2015 09:53:50 -0800

ariddell:
Thanks. (By the way, I think dnsmasq was added so that dns requests could be routed via the server, see #236)

roles/vpn/templates/etc_openvpn_server.conf.j2
```diff
@@ -298,9 +299,18 @@ status openvpn-status.log
...
 # 9 is extremely verbose
-verb 3
+verb {{ openvpn_verb }}
 
 # Silence repeating messages.  At most 20
 # sequential messages of the same message
 # category will be output to the log.
 ;mute 20
+
+# Mask your settings with mtu
```
ariddell: Would you mind adding a bit more explanation, like "Your mtu can be used by others to detect VPN use", "Mask your settings by using a lower mtu".

Thanks again for this patch.

Friz-zy:
I uses pure openvpn with pushing google dns and no one additional options on client and dnsleaktest.com detects google server =)

Friz-zy:
[Russian post](http://habrahabr.ru/post/216295/)
tldr: openvpn changes length of network packages in the certain chunk that depends on chipher and hash-sum algorithms

roles/vpn/templates/etc_openvpn_server.conf.j2
Friz-zy: [Russian post](http://habrahabr.ru/post/216295/)
tldr: openvpn changes length of network packages in the certain chunk that depends on chipher and hash-sum algorithms

Friz-zy:
Thanks!

neuhaus:
Ubuntu Trusty comes with openvpn 2.3.2 btw

Friz-zy:
Nice work ;)
I have another additions for you ;)
I'll send a request later

jplock:
Per my comment on #439, the default MTU is causing problems for me with iOS devices.  Can we change the default to something other than `1300`?  I'm running Debian Wheezy.

```
root@mail:/etc/openvpn# dpkg -l | grep openvpn
ii  openvpn                               2.2.1-8+deb7u3                    amd64        virtual private network daemon
```

Friz-zy:
I found an [old bug](https://community.openvpn.net/openvpn/ticket/334): looks like ios client doesn't understand *-mtu options. Unfortunately I don't have an ios for tests...
I think in your case you can only disable it.
@ariddell, should we add a comment into config about IOS?

ariddell:
The client doesn't understand link-mtu in the .ovpn, right? Can you just set it manually? Or does iOS refuse to let you set the MTU at all?

I suppose adding a comment wouldn't hurt. Perhaps someone can confirm that you can manually set the MTU on iOS?

(It will be great to disaggregate these roles since one could add this information to the README of the vpn role.)

jplock:
Can we comment out the MTU option by default and let people uncomment it if they they'd like to?

ariddell:
I feel odd about accommodating a broken iOS client. What could be more basic than setting a lower MTU?

neuhaus:
From what I understand, we used to have no default MTU set in the server, why don't we just remove the default MTU again?

Friz-zy:
It's easy to detect openvpn traffic by mtu fingerprint (even cipher and hash sum algorithm), look at witch.valdikss.org.ru
We can use msfix instead but I couldn't get it well configured and it still can be invalid for ios...

ariddell:
I've had to lower the MTU in other contexts -- it's a very normal
operation. Is there any clue as to why the iOS client doesn't support
it?

I suppose this comes down to privacy (i.e., fingerprinting resistance)
vs. iOS support. While I'm likely somewhat biased in that I don't use
iOS, the fingerprinting resistance seems like a sensible change and very
much in the spirit of sovereign.

Friz-zy:
Myabe we can add new option or variable as a solution?
Something like 'disable this if you uses ios clients, default is Enabled'.
I don't have any information about ios client except google search but I find this situations at least strange too...

ariddell:
We're going to add a variable to account for a broken client? I'd prefer to leave the fingerprintability for now than go down that route.

neuhaus:
Fine with me - add the MTU to the configuration and make sure to mention it in the README and in a comment in the config file that it is incompatible with iOS.

Friz-zy:
I agree with @neuhaus

ariddell:
So it will be a config option that just specifies the MTU  -- but if
it's None it removes the link-mtu line entirely?

neuhaus:
Sounds good, if we add a variable, it will help users that want to work with iOS because their changes won't be overwritten all the time.

ariddell:
Great, @Friz-zy, will you submit a PR?

Friz-zy:
Yep, I'll do it today later

Friz-zy:
Pushed to https://github.com/sovereign/sovereign/pull/445

roles/vpn/templates/etc_openvpn_server.conf.j2
```diff
@@ -188,6 +188,7 @@ ifconfig-pool-persist ipp.txt
 # or bridge the TUN/TAP interface to the internet
 # in order for this to work properly).
 ;push "redirect-gateway def1 bypass-dhcp"
+;push "dhcp-option DNS 8.8.8.8"
```
neuhaus: we removed this line earlier, don't want to promote services run by 3rd parties unless necessary.

vars/defaults.yml
neuhaus: Ubuntu 14.04.03 ships with OpenVPN 2.3.2.

Issue #190, opened 2014-03-19
Roundcube reports "Connection to imap server failed"


jgillman:
I just updated my SSL cert as the first time I made it I wasn't as knowledgable and improperly created it. After installing them, I restarted my server (Debian on Linode), and everything seemed to come back up fine, but now when I navigate to Roundcube it reports "connection to imap server failed"

Apache is correctly using the new SSL cert. I'm not really sure where to start trouble shooting this.

Issue #412, opened 2015-09-21
Modifying public / private keys results in system not working after ansible re-run


shaneonabike:
Honestly, I cannot wait until Let's Encrypted releases their public keys for everyone life is going to be so much easier for everyone.

But until then... I seem to be having an issue I was wondering if I could get some help on.

Specifically, I signed my own keys and now for some reason both dovecot and postfix are not happy campers.

dovecot: Error: service(imap-login): command startup failed
postfix/smtpd : warning: cannot get RSA priate key from file

Sep 21 19:31:55 localhost postfix/smtpd[10619]: warning: TLS library problem: error:0B080074:x509 certificate routines:X509_check_private_key:key values mismatch:x509_cmp.c:330:

I read that this can be an issue relating to self-signed because they do no include 'localhost' and therefore are not working properly. Should I be modifying something to make this work? I cannot get my Roundcube or Thunderbird to connect....

MisterWS:
Heya - I got this issue a few weeks ago when I changed up my certificates and keys. In the end, it turned out to be because the following line in common/roles/security.yml doesn't know it needs to update the combined wildcard_combined.pem file:

- name: Create a combined version of the public cert with intermediate and root CAs
  shell: cat /etc/ssl/certs/wildcard_public_cert.crt /etc/ssl/certs/wildcard_ca.pem >
    /etc/ssl/certs/wildcard_combined.pem creates=/etc/ssl/certs/wildcard_combined.pem

I corrected by manually deleting wildcard_combined.pem from the server and re-running. And then completely forgot to raise a bug report for it. Oops.

shaneonabike:
BAM! Genius.. I have yet to check if this combine certificate exists elsewhere but that was my problem!

PR #430, opened 2015-11-19
sl: properly install changed certificate and restart apache on changes

danmilon:
I was stupidly using the certificates provided in the repo, and then I generated some self-signed ones but found out that they were not properly installed. The `creates=/etc/ssl/certs/wildcard_combined.pem` was making the task to skip which it shouldn't. So I fixed that, and also thought it should restart apache by itself.


neuhaus:
Can you please make a separate PR for the SSL stapling thing?

danmilon:
See #434

danmilon:
I also removed the SSL stapling from this PR.

neuhaus:
fixes #190, #412 and #449

Issue #426, opened 2015-11-05
Create SSL certificates using Let's encrypt


neuhaus:
[Let's encrypt](https://letsencrypt.org) is a new free and automated certificate authority that will enter productive mode this month.

I plan to add support for it to sovereign to replace the need for an expensive wildcard certificate, however it will be quite a bit of work so any help is appreciated.

masterkorp:
The link is wrong, correct one is https://letsencrypt.org/

tracker1:
+1 was coming in to suggest the very same...  Will be very cool to have automated letsencrypt SNI setup for the custom domain(s) in question...

neuhaus:
Fixed the link. So, do you guys want to help? :-) Got suggestions regarding which client to use? There is the official one of course (written in python), but there are also alternatives.

tracker1:
If python is already in the dependencies, I'd just go with the default/official client.

sread:
Did a bit of research, since we're not the only ones interested in this. thefinn93 ( https://github.com/thefinn93/ansible-letsencrypt ) ansible role looks like a good starting point, see also ( https://community.letsencrypt.org/t/completely-automated-setup/3121/2 )

neuhaus:
The official client is not very popular because it tries to do too many things.
Here are some simple alternatives:
* https://github.com/kuba/simp_le
* https://github.com/diafygi/acme-tiny/
* https://github.com/xenolf/lego *(needs golang)*
* https://github.com/barosl/letsencrypt-simple
* https://github.com/hlandau/acme.t *(needs golang)*

danmilon:
What do you think the overall architecture should be? I can try to tackle this. All needs to happen in the host where things are being provisioned, right?

One problem I thought of, is that you don't know which roles are activated, thus which domains to ask the certificate for. One solution to that is one certificate for each service.

neuhaus:
The DNS part is outside of the scope of the sovereign project. I.e. the user has to make sure the DNS records are in-place. The playbook should make sure to check a DNS record before trying to create a Let's encrypt certificate.

danmilon:
I'm talking about [SANs](https://en.wikipedia.org/wiki/SubjectAltName). When you ask LetsEncrypt for the cert, if you're going to install one cert for all services, you need to ask for which subdomains the certificate will be valid. LetsEncrypt does not support wildcard certificates.

And there will be a validation for each subdomain you ask, so it needs to work, cant be arbitrary.

[...]

PR #431, opened 2015-11-19
Optionally add HSTS header

danmilon:
For self signed certificates, when the server adds the HSTS header, firefox will refuse to connect to it, and won't even let you add to add an exception. So I added a variable that by default adds the header.

I am not sure how well we can communicate the existence of this option with the users. Should I add it in the README?

### DIFF
--- a/roles/common/templates/etc_apache2_ssl.conf.j2
+++ b/roles/common/templates/etc_apache2_ssl.conf.j2
@@ -11,4 +11,7 @@ SSLCipherSuite ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-R
 SSLCertificateFile      /etc/ssl/certs/wildcard_public_cert.crt
 SSLCertificateKeyFile   /etc/ssl/private/wildcard_private.key
 SSLCACertificateFile    /etc/ssl/certs/wildcard_ca.pem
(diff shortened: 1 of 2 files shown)


danmilon:
ping, anyone there?

neuhaus:
If you add a new option, it needs to be documented.
How do other browsers behave with regards to self-signed certificated and the HSTS header?

danmilon:
I tried with chromium and it allows you to override even with HSTS and self signed. I believe we should work on lets encrypt certs instead of self signed, so we wont need this at all. What do you think?

neuhaus:
Agreed. See issue #426

nodiscc:
I can't reproduce this, Firefox (Iceweasel 42.0-1) allows me to override with self-signed cert AND HSTS headers (it just doesn't enforce HSTS).

danmilon:
Indeed now that I put back HSTS I cant reproduce it also. There are some bugs reported for firefox (e.g. [800882](https://bugzilla.mozilla.org/show_bug.cgi?id=800882)). Since I fixed the issue by editing [SiteSecurityServiceState.txt](http://superuser.com/questions/539580/how-can-i-remove-a-website-accidentally-added-to-firefoxs-list-of-hsts-sites#comment1276327_912277), noone else has reported it, and I can't reproduce it, I'll close this.

danmilon:
For reference, see this screenshot of firefox 42.
screen shot 2015-12-08 at 19 45 45

PR #433, opened 2015-11-20
Fix selfoss permissions

danmilon:
I got this mail from cron/logrotate:

```
/etc/cron.daily/logrotate:
error: skipping "/var/www/selfoss/data/logs/default.log" because parent directory has insecure
permissions (It's world writable or writable by group which is not "root")
Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
run-parts: /etc/cron.daily/logrotate exited with return code 1
```

i think the permissions I set in this PR are more appropriate.

### DIFF
--- a/roles/news/tasks/selfoss.yml
+++ b/roles/news/tasks/selfoss.yml
@@ -9,7 +9,11 @@
 
 # only data/cache, data/favicons, data/logs, data/thumbnails, data/sqlite public/ should be writeable by httpd
 - name: Set selfoss permission
-  action: file path=/var/www/selfoss/{{ item }} mode=0775
+  action:
+    file path=/var/www/selfoss/{{ item }}
+    owner=www-data
+    group=www-data
+    recurse=yes
   with_items:
     - data/cache
     - data/favicons


danmilon:
With these permissions, I get the same error.
```
[removed]@[removed]:/var/www/selfoss/data/logs$ ls -la
total 1468
drwxrwxr-x 2 root www-data    4096 Dec  2 11:25 .
drwxr-xr-x 8 root www-data    4096 Nov 18 09:25 ..
-rw-rw-r-- 1 root www-data 1339221 Dec  4 11:55 default.log
-rw-rw-r-- 1 root www-data   30577 Dec  2 11:25 default.log.1.gz
-rw-rw-r-- 1 root www-data   28480 Dec  1 11:25 default.log.2.gz
-rw-rw-r-- 1 root www-data   26159 Nov 30 11:25 default.log.3.gz
-rw-rw-r-- 1 root www-data   25277 Nov 29 11:25 default.log.4.gz
-rw-rw-r-- 1 root www-data   27408 Nov 28 11:25 default.log.5.gz
-rw-rw-r-- 1 root www-data      13 Nov 18 09:25 .htaccess
```
```
/etc/cron.daily/logrotate:
error: skipping "/var/www/selfoss/data/logs/default.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
run-parts: /etc/cron.daily/logrotate exited with return code 1
```

`Writeble by a group which is not "root"` is what it doesn't like probably.
Not sure why it worked when both user and group is `www-data`.
Is there any reason why we dont want it that way?

neuhaus:
If there is a successful attack on the webserver, the attacker can run things as user `www-data`.
Stuff owned by www-data can then be modified, such as permission on a directory owned by www-data.
If the user www-data merely has write permissions on a directory, he cannot change its permission.

neuhaus:
Could you please try the suggested fix, adding a "su" line to `/etc/logrotate.d/selfoss`copied from `etc_logrotate_selfoss`?

Like this:
   su www-data www-data

danmilon:
You are right, it worked like that. If you believe that from a security standpoint its OK, I can send a PR.
```
root@srv1:/var/www/selfoss/data/logs# ls -la
total 1544
drwxrwxr-x 2 root www-data    4096 Dec  4 14:45 .
drwxr-xr-x 8 root www-data    4096 Nov 18 09:25 ..
-rw-rw-r-- 1 root www-data    3189 Dec  4 14:45 default.log
-rw-rw-r-- 1 root www-data 1411969 Dec  4 14:40 default.log.1
-rw--w---- 1 root www-data       0 Dec  4 14:44 default.log.1.gz
-rw-rw-r-- 1 root www-data   30577 Dec  2 11:25 default.log.2.gz
-rw-rw-r-- 1 root www-data   28480 Dec  1 11:25 default.log.3.gz
-rw-rw-r-- 1 root www-data   26159 Nov 30 11:25 default.log.4.gz
-rw-rw-r-- 1 root www-data   25277 Nov 29 11:25 default.log.5.gz
-rw-rw-r-- 1 root www-data   27408 Nov 28 11:25 default.log.6.gz
-rw-rw-r-- 1 root www-data      13 Nov 18 09:25 .htaccess
```
```
root@srv1:/var/www/selfoss/data/logs# cat /etc/logrotate.d/selfoss 
/var/www/selfoss/data/logs/*.log {
        daily
        missingok
        rotate 5
        compress
        notifempty
        su www-data www-data
#	create 0644 www-data www-data
}
```
```
root@srv1:/var/www/selfoss/data/logs# logrotate /etc/logrotate.d/selfoss 
```

danmilon:
See #438.

neuhaus:
Merged #438

Issue #406, opened 2015-08-16
SSLUseStapling On in apache config cause problems


hippich:
It appears that debian wheezy has apache 2.2.x in their repositories and SSLUseStapling option as available starting 2.4.x.

neuhaus:
Confirmed. For now, please upgrade to Apache 2.4 or remove that option.

shaneonabike:
I added a wiki page for this to help people out. Maybe we could add something to the README also just to flag this or at least tell people to head to the [wiki](https://github.com/sovereign/sovereign/wiki/Apache-Stapling-(Compatibility-only-with-Apache-2.4)) if they are having issues. 



@hippich do you think you could provide more info on the wiki if people want to upgrade their apache. I want to do that but was curious as to whether I needed to upgrade my PHP also. I suspect not right?

hippich:
@shaneonabike, i just removed SSLUseStapling from my configs.

sread:
It seems that based on this [page](https://www.digitalocean.com/community/tutorials/how-to-configure-ocsp-stapling-on-apache-and-nginx) we need to have an SSLStaplingCache:

   SSLStaplingCache shmcb:/tmp/stapling_cache(128000)

and this needs to be defined outside  <VirtualHost></VirtualHost> so it can't go in the existing ssl.conf includes. A conf-available/ssl-stapling-cache.conf file ((a2enconf enabled) worked for me (although the actual stapling fails due to the self-signed cert in the vagrant box).

danmilon:
What @sread said. I run into this issue also. @al3x what is the proposed sollution for jessie? I can send a PR.

neuhaus:
Debian Jessie comes with Apache 2.4 so SSLUseStapling works there.
We should remove it for Debian 7 because it breaks the default.

neuhaus:
Wait, isn't this already fixed since July 16th in commit 1be1afe?

It explicitly checks for wheezy:

    {% if ansible_distribution_release != 'wheezy' %}
        SSLUseStapling On
        SSLStaplingResponderTimeout 5
        SSLStaplingReturnResponderErrors off
    {% endif %}

Please clarify otherwise I will consider this issue to be fixed.

sread:
I believe there are two problems: 

1. Apache < 2.3.3 (fixed as you say)
2. Lack of defined SSLStaplingCache causes apache to fail to start

The confusion is due to my mistake of hijacking the bug (same result, different cause)

[...]

PR #434, opened 2015-11-24
Add SSL stapling cache for apache

danmilon:
See #406.


danmilon:
force pushed. please review again.

PR #436, opened 2015-12-02
Remove reference to Google's DNS servers

ariddell:
Per discussion on #429 (after the merge). This project is about encouraging users to run services themselves and not rely on for-profit corporations such as Google.


neuhaus:
Thanks

PR #439, opened 2015-12-05
update openvpn server config

Friz-zy:
Change default network buffer size
Should increase tcp tunnel speed for openvpn < 2.3.9
https://community.openvpn.net/openvpn/ticket/461

### DIFF
--- a/roles/vpn/templates/etc_openvpn_server.conf.j2
+++ b/roles/vpn/templates/etc_openvpn_server.conf.j2
@@ -315,3 +315,11 @@ tun-mtu {{ openvpn_mtu }}
 # Only for openvpn 2.3.3 and >2.3.4
 {{ openvpn_tls_version_min }}
 {{ openvpn_tls_cipher }}
+
+# Change default network buffer size
+# Should increase tcp tunnel speed for openvpn < 2.3.9
+# https://community.openvpn.net/openvpn/ticket/461
+sndbuf 0
+rcvbuf 0
+push "sndbuf 393216"
+push "rcvbuf 393216"


ariddell:
Thanks for this! Would you mind adding a note about which version of openvpn has this included so we can know when it's no longer needed (perhaps the release after Jessie)?

ariddell:
@Friz-zy on another note, do you have a working openvpn config for jessie? We're missing that.

Friz-zy:
Even unstable sid right now [uses 2.3.7](https://packages.debian.org/search?keywords=openvpn) so it will be not so soon...
About config: what's the problem?
Unfortunately, I am using an older version of Ubuntu with not the most fresh openvpn.

ariddell:
But there is a release which includes the fix, right? We should note that.

re: config, jessie uses systemd, so a few things need to change.

Friz-zy:
Fix will be in openvpn 2.3.9, but I don't know what distro will include it.
With systemd you have a problem with rc.local, haven't it?
I can fix it by rc-local.service, if you want.

ariddell:
Oh, I'm sorry -- I didn't realize you included the version number in the
first place.

jplock:
I can no longer connect to my VPN from my iPhone or iPad after this change.  Desktop works fine though.  Any thoughts?

Friz-zy:
Hm, can you change
```
push "sndbuf 393216"
push "rcvbuf 393216"
```
to
```
+push "sndbuf 0"
+push "rcvbuf 0"
```
or delete it and post results here?

jplock:
So after some trial and error, this PR is actually fine.  The problem I'm running into is with the `openvpn_mtu` change in https://github.com/sovereign/sovereign/pull/429.  I had to comment that out to be able to connect.  I'm running Debian Wheezy.

```
root@mail:/etc/openvpn# dpkg -l | grep openvpn
ii  openvpn                               2.2.1-8+deb7u3                    amd64        virtual private network daemon
```

PR #445, opened 2015-12-10
vpn: add systemd compatibility for rc.local

Friz-zy:
I don't have system for check it right now but it should help.

### DIFF
--- a/roles/vpn/files/rc-local.service
+++ b/roles/vpn/files/rc-local.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=/etc/rc.local Compatibility
+After=network.target
+
+[Service]
+Type=oneshot
+ExecStart=/etc/rc.local
+TimeoutSec=0
+StandardInput=tty
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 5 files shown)


ariddell:
Thanks so much for your work on this.

Let's merge the systemd stuff into the jessie branch and split the mtu fix out.

Is there any ansible pattern for "if X is defined, then Y"? It would be nice to avoid putting the 'tun-mtu' in the variable name itself.

Friz-zy:
Sorry, I want split two pull requests but github automatically pushed second commit to opened PR =\

Yes, Ansible has this pattern:
```
- debug: msg="Defined!"
  when: variable is defined
```

ariddell:
So can we do something like ``{% if variable is defined %}tun-mtu {{ variable }} {% endif %}``?

Can we be more explicit about the instructions for iOS users? Can they set the variable to ``None`` or something instead of commenting the line out?

If you don't want to deal with github's PR system I'm happy to split out your commits (when they're done) and cherry-pick them into separate pull requests.

Friz-zy:
You gave me good advice with cherry-pick ;)
I'll split it to two different PR.

Yep, It's a good point: because it's a jinja template we can check if variable exists and not false\empty. Nice solution!

ariddell:
This still needs some work. @Friz-zy you mentioned you would split this into different PRs. Looking forward to them.

Issue #425, opened 2015-11-04
Apache doesn't serve http redirect


lukecyca:
I'm finding that my blog isn't accessible via http, only via https. The intended behaviour is to issue a redirect to https, as well as a `Strict-Transport-Security` to tell the browser to use https for subsequent requests to the hostname. I suspect this hid the issue from me for a long time.

To test your own site: `curl -v example.com`. Expected behaviour is to get a 301 redirect, but actual behaviour is a 403 Forbidden.

I've isolated the issue to the status vhost used for monitoring apache, originally added last year by @philandstuff. If I disable that vhost, it works as intended again.

I have a handful of other local modifications, so I'd appreciate it anyone else can try reproduce this issue. This could be the issue that @almereyda was describing in #250.

neuhaus:
Interesting. Can you run Apache in debug mode (parameter -X)?

iboxifoo:
I found the solution at https://github.com/sovereign/sovereign/issues/360 fixed this particular issue.

neuhaus:
Does changing fqdn.conf to "ServerName localhost" also fix the issue for you, @lukecyca ?

lukecyca:
This is indeed a duplicate of #360.

Issue #360, opened 2015-02-24
fqdn.conf ServerName matching ServerName in VirtualHost can mess up HTTP (not HTTPS)


ariddell:
Apparently under trusty if you have
```
ServerName mydomain.org
```
in ``/etc/apache2/conf-available/fqdn.conf``

and also have ``ServerName mydomain.org`` in your apache2 configuration for the blog you can
screw up your *http://mydomain.org* (https will still work). I fear people might not notice this because of HTST pinning.

This might only show up in edge cases (I have an additional configuration in `sites-enabled`), but the error message is no help. It's:

```AH01797: client denied by server configuration: /var/www/```

The problem is fixed by putting anything else (e.g., "localhost") in fqdn.conf.

I think I have this right. Perhaps an Apache guru could explain what happened and why the fix works.

**Comment by cjchacon** [comment]
Thanks, I suffered this issue too, and solved it with your suggestion. I found the file in `/etc/apache2/conf.d/fqdn` and put `ServerName localhost`.

**Comment by neuhaus** [comment]
Will putting only "localhost" in fqdn.conf make bug #187 reappear?

**Comment by lukecyca** [comment]
Confirmed that setting `/etc/apache2/conf.d/fqdn` to `ServerName localhost` solves this problem (as well as #187) for me.

**Comment by mikeashley** [comment]
I wonder if Apache still has the problem originally described in this issue?  I was not able to reproduce it, and I set my fqdn to `ServerName {{ domain }}`.  @ariddell Do you know what version of Apache you were running when you reported this?

**Comment by mikeashley** [comment]
I'll consider this "cannot reproduce."  Please reopen if we find this problem on the current master branch.

**Comment by lukecyca** [comment]
This issue just bit me again after upgrading to jessie (`master` branch).

The suggested workaround (changing `fqdn.conf` to `ServerName localhost`) fixed it as before.

**Comment by lukecyca** [comment]
This is _still_ an issue for me. Every once in a while an upgrade reverts my fix, and I go a few months with a broken website (it works for _me_ because HTST).

My local workaround is:
```diff
diff --git a/roles/common/templates/fqdn.j2 b/roles/common/templates/fqdn.j2
index 213beec..873caeb 100644
--- a/roles/common/templates/fqdn.j2
+++ b/roles/common/templates/fqdn.j2
@@ -1 +1 @@
-ServerName {{ domain }}
+ServerName localhost
```

I'm on Debian Jessie.

neuhaus:
[...]

PR #451, opened 2015-12-27
LetsEncrypt support

mikeashley:
This is a draft pull request to advance the discussion in issue #426.  I am interested in feedback on the approach and recommendations on how to improve it.  It is incomplete but is enough to see how it will work.  Known issues:

1. stopping/starting apache is awkward, especially if you don't really need a certificate; and
2. no story for certificate renewal.

What else am I missing? Also, I am an Ansible rookie, so suggestions on how to code this better are welcome.

I am using the official client since LetsEncrypt is still in beta.


ariddell:
This looks really good. Could we create a new role ('acme') rather than putting more stuff in common? We can run `acme` right after common.

As I'm sure you're aware, with let's encrypt's 3 month expiration I think automatic renewal has to be in. I guess we need a cron job?

Thanks again for this.

mikeashley:
Allen, thank you for the encouragement.  I agree using cron is probably the way to go for renewing certificates.

I am not sure a separate role makes sense.  That would introduce a new dependency for modules that need certificates.  Can that be managed automatically without risk of a misconfigured site.yml?

ariddell:
One thing to check about the renewal step is if the relevant apps/services need to be restarted to use the new certificate. I suspect they will all need to be restarted.

re: acme role: Having a separate role would increase the readability of the code. That it's a mandatory role (like ``common``) is just something we need to state explicitly. Handling dependencies is clearly not something that ansible does very well.

mikeashley:
My apologies.  I'm still figuring out the right way to use GitHub's pull request workflow.  I'm moving the second draft pull request from #455 back to this pull request to consolidate.

This is a different approach to using LetsEncrypt to replace the wildcard certificate we currently use. In contrast to what I originally proposed, this generates only one certificate using Subject Alternative Names to list the subdomains that Sovereign roles may use. Since we are not using a wildcard certificate, roles are always going to have their fingers in certificate generation. Given that, I went for broke and listed subdomains explicitly.

Based on feedback above, I put the LetsEncrypt stuff in a separate task list but elected to keep it in the common role. It just doesn't seem warranted to put it in a separate role. Also, I added a cron script to renew the certificate periodically (once a month in this draft).

mikeashley:
Note: Allen suggested making the letsencrypt email optional instead of required.

ariddell:
e-mail is optional in the ACME spec (https://ietf-wg-acme.github.io/acme/), which I think should be considered authoritative relative to some comments on a forum.

We have a private registration key. It's backed up to tarsnap (right?). The email would only be used if we lose access to the private key. But even in that case we likely wouldn't need it anyway since we presumably control the DNS we can just re-register with let's encrypt. I can't figure out any situation where providing the email to let's encrypt would be *desirable* in the present context, much less mandatory.

mikeashley:
Good catch on keys backup!

You're right, the forum thread doesn't mean too much, but I pointed to it since this topic does come up on their forums.  I don't know how to reconcile "optional in the spec" with the standard client's strong language when the address is omitted.

Examples of when it would be useful for LE to notify you: major changes in their service and security problems with their certificates.

For the target user of Sovereign, what's the right way to default this: require the email address or make it optional?  I think expecting an email address is the right default since that is what the LE team wants.  Do we not trust them with an email address?

Finally, I don't personally feel that strongly about this.  I can go either way.  I'm just looking for some consistent design intent from the Sovereign maintainers so that I know how to make decisions like this when putting together roles.

ariddell:
Hopefully someone else can cast their vote here. I only feel strongly about it not being *required*; it's just contact information that the user should feel free to provide or not (that much is clear from the spec). If the email isn't provided the thing should still work.

I'm happy to help with any of these final details.

mikeashley:
There is a bug with this design.  It requires the user to add DNS records for all subdomains whether or not she will use them.  This is because LE will look up each subdomain listed in the request.

ariddell:
Any chance it can fail gracefully? That is, if the user didn't add a record for ``news.domain`` or ``read.domain``, it just doesn't issue a certificate? (or creates something signed by the ``domain`` private key?)

mikeashley:
Tests are passing on wheezy and trusty.  I still need to fix up the implementation of certificate renewal.

jplock:
Can we leave the readme instructions for using your own certificate in place?

mikeashley:
With this approach, supporting both LE certificates and your own wildcard certificate will take some work.  I'm wondering how many people would buy a certificate or use a self-signed certificates are available at no cost from LE.  Thoughts on this?

jplock:
For me, I'm using a wildcard cert I just renewed for a few years. If this gets merged, it's going to override that certificate everywhere. I'm just saying it would be nice to try and phase this in optionally. If others don't feel the same way, that's fine. I'll just have to maintain my own fork.

ariddell:
Perhaps we could introduce this in the Jessie branch since people will
be expecting things to break during that transition?

I think LE is important and we should definitely make it the default.

Also -- what's wrong with just switching over to LE for your TLS
certificate? LE will issue a certificate even if you have a certificate
from another CA, right?

mikeashley:
I like Allen's idea: introduce it in the Jessie branch.  I am currently running Wheezy, and my purchased certificate expires in February.  I would need to keep this work branched from master also unless the Jessie branch is close to ready.  Maybe make the branch from master available for the adventurous who don't want to buy a certificate.

jplock:
Agreed. I like the idea of introducing it on the jessie branch. I'm still on wheezy.

mikeashley:
I rebased to catch up to master and resolve a conflict.  I also took the opportunity to clean up the branch to make it easier to review.

README.md
```diff
@@ -168,6 +144,22 @@ For Git hosting, copy your public key into place:
 
 Finally, replace the TODOs in the file `hosts`. If your SSH daemon listens on a non-standard port, add a colon and the port number after the IP address. In that case you also need to add your custom port to the task `Set firewall rules for web traffic and SSH` in the file `roles/common/tasks/ufw.yml`.
 
+### 4. Set up DNS
+
+If you’ve just bought a new domain name, point it at [Linode’s DNS Manager](https://library.linode.com/dns-manager) or similar. Most VPS services (and even some domain registrars) offer a managed DNS service that you can use for this at no charge. If you’re using an existing domain that’s already managed elsewhere, you can probably just modify a few records.
+
+Create `A` records which point to your server's IP address:
```
neuhaus: LE is working on IPv6 certification (see https://github.com/letsencrypt/boulder/issues/593)

README.md
```diff
@@ -61,31 +61,7 @@ What You’ll Need
 Installation
 ------------
 
-### 1. Get a wildcard SSL certificate
```
neuhaus: There should be at least one sentence that explains that trusted SSL certificates will be be provided using Let's Encrypt.

roles/common/DESIGN.md
```diff
@@ -0,0 +1,49 @@
+# Design Description for Common Role
+
+## Let's Encrypt Support
```
neuhaus: Not everyone will be familiar with Let's encrypt, we need a short introduction ("Let's Encrypt is an automated certificate authority providing SSL certificates that are trusted by all major browsers").

roles/common/templates/etc_apache2_ssl.conf.j2
```diff
@@ -2,13 +2,13 @@ SSLEngine on
 SSLProtocol ALL -SSLv2 -SSLv3
 SSLHonorCipherOrder On
 SSLCompression off
+SSLCertificateKeyFile	/etc/letsencrypt/live/{{ domain }}/privkey.pem
+SSLCertificateFile	/etc/letsencrypt/live/{{ domain }}/cert.pem
+SSLCertificateChainFile	/etc/letsencrypt/live/{{ domain }}/chain.pem
```
neuhaus: Mobile Safari will complain about the certificate unless we use fullchain.pem here. See also: https://community.letsencrypt.org/t/untrusted-on-mobile-newbe/7903

roles/xmpp/tasks/prosody.yml
```diff
@@ -37,6 +34,23 @@
...
+    - privkey.pem
+    - cert.pem
+
+- name: Assert mode and ownership on SSL private key and cert
+  file: dest=/etc/prosody/certs/{{ item }} owner=root group=prosody mode=0640
+  with_items:
+    - privkey.pem
+    - cert.pem
+
+- name: Update certificate renwal cron job
```
neuhaus: typo renewal

roles/xmpp/tasks/prosody.yml
```diff
@@ -37,6 +34,23 @@
   template: src=prosody.cfg.lua.j2 dest=/etc/prosody/prosody.cfg.lua group=root owner=root
   notify: restart prosody
 
+- name: Copy SSL private key and cert
```
neuhaus: Why can't prosody use the same file?

roles/common/tasks/letsencrypt.yml
```diff
@@ -0,0 +1,83 @@
...
+    dest=/etc/cron.monthly/letsencrypt-renew
+    owner=root
+    group=root
+    mode=755
+
+- name: Create live directory for LetsEncrypt cron job
+  file: state=directory path=/etc/letsencrypt/live group=root owner=root
+
+- name: Stop Apache
+  service: name=apache2 state=stopped
```
neuhaus: Invoking letsencrypt with -a webroot  and "certonly" would allow us to leave Apache running.

roles/common/tasks/letsencrypt.yml
```diff
@@ -0,0 +1,83 @@
...
+
+- name: Get an SSL certificate for {{ domain }} from Let's Encrypt
+  command: /root/letsencrypt/letsencrypt-auto certonly -c /etc/letsencrypt/cli.conf --domains {{ domain }},{{ subdomains }}
```
neuhaus: As mentioned earlier, not every user will use every role but currently this would force her to create all the DNS entries. I'd prefer a small shell function that does the DNS lookups and creates a list of all subdomains that have working DNS.

roles/common/files/etc_cron-monthly_letsencrypt-renew
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# Renew all live certificates with LetsEncrypt.  This needs to run at least
+# once every three months.
+
+# Given a certificate file returns "domain1,domain2"
+# https://community.letsencrypt.org/t/help-me-understand-renewal-config/7115
```
neuhaus: cool

ariddell:
Just a reminder: this PR is currently going to be merged into master (rather than the jessie branch). It seems like there's agreement it should be merged into the jessie branch.

When this is done I'm ready to test it + the jessie branch. Thanks for all your work on this @mikeashley.

neuhaus:
How do you feel about [using a more minimal ACME client](https://github.com/sovereign/sovereign/issues/426#issuecomment-161972594)? I'm not a big fan of the bloated standard client myself.

mikeashley:
I started with the official client since LE is in beta.  If there is a simpler client that does what we want directly, we should switch.

mikeashley:
@neuhaus I appreciate the detailed review.  Thank you!  Next push of this PR will be based on the jessie branch unless there are objections.

roles/common/tasks/letsencrypt.yml
mikeashley: As it stands, the user wouldn't need to create the DNS entries. They would override this variable in their user.yml setup and list only what they are using.

Still, I have never liked this because it's an opportunity for a configuration mistake.  I like your idea.  Do you know what the basic mechanic would be for sniffing out the actual domains with working DNS?

roles/xmpp/tasks/prosody.yml
mikeashley: Permission problems.  The server runs as user/group prosody.  This could be solved with a group to access /etc/letsencrypt, but I wasn't comfortable whacking permissions on /etc/letsencrypt since I don't know what the client does after initial setup.

This is something to consider when we're evaluating other clients.

README.md
mikeashley: Sorry, I'm not following what change you are proposing here.

roles/xmpp/tasks/prosody.yml
neuhaus: But if the user prosody is in the group ssl-cert it can access the files...

roles/common/tasks/letsencrypt.yml
neuhaus: We can use `/usr/bin/host from the `bind9-host` package. It returns exit code 1 if the hostname doesn't resolve and 0 if it resolves ok.

roles/xmpp/tasks/prosody.yml
mikeashley: Not as it stands.  The LE client sets up /etc/letsencrypt with root.root ownership.  We'd have to whack at least a portion of the tree recursively to root.ssl-cert.

roles/common/tasks/letsencrypt.yml
neuhaus: And finally there is `/usr/bin/getent` from the `libc-bin` package that should be installed everywhere.
`getent ahosts <hostname>` returns error code 2 if the name doesn't resolve.

roles/xmpp/tasks/prosody.yml
neuhaus: OK, now I understand. Changing the group ownership to ssl-cert seems like a good idea though, doesn't it?
There's already an open issue for it at https://github.com/letsencrypt/letsencrypt/issues/1425

mikeashley: Yes, it's a good idea.  The open issue at letsencrypt is exactly right.  I'll go ahead and make the change to fix permissions and the corresponding update to the prosody role to avoid copying the files.

mikeashley:
These seven commits address feedback from @neuhaus.  I haven't tried invoking the LE client in a way that avoids having to stop Apache yet.

ariddell:
I don't know how you change the target of the PR. Is it even possible?

I take it this is ready for testing? I'll plan on testing it out sometime this week.

mikeashley:
Good question.  Rebasing definitely seems to have confused github.  I rebased onto commit 89d0c076592820f91bc3f6c4759ef1d56e4b4127 (currently the head of the jessie branch).

I deployed what I had based on master to my own server several days ago, and it's working fine.  I smoke-tested this branch on a Debian 8.1 server, and it looked ok.  It needs more testing at this point.

The jessie branch still seems a little rough.  I had to disable z-push, because the task failed.  ZNC also failed to run, because the task assumes a group and owner called 'znc', but the jessie branch switches to an apt package that installs as root.  In short, you will probably need to fiddle.  I definitely recommend starting with a minimal set of roles.

mikeashley:
I took a brief look at running the LE client with webroot authentication.  To do it will require running the gauntlet of virtual host configurations various roles set up.  My guess is that Apache is flexible enough to do what needs done, but it will take some homework to figure out the right changes to the virtual host configurations.  My understanding of Apache configuration is pretty basic.  Somebody fluent could probably knock it out much faster than me.

neuhaus:
My expectation is that with the standard setup we will have multiple virtualhosts with different documenroot only on port 443. On port 80 we have one Apache instance that will answer for all hostnames, deal with requests for everything starting with `/.well-known/` by itself and redirect to the HTTPS services for all other requests. 
If you need help with that let me know.

mikeashley:
That is my thinking as well.  Help would be appreciated.  I am traveling next weekend so may not be able to spend time on this until the weekend of Jan 30.  If nobody has made progress, I'll take a swing at it then.

ariddell:
On a fresh Jessie install I'm hitting a SSLStapling log. I think this might have been fixed somewhere else.

```
[Tue Jan 19 00:14:44.414523 2016] [ssl:emerg] [pid 26592] AH01958: SSLStapling: no stapling cache available
[Tue Jan 19 00:14:44.414567 2016] [ssl:emerg] [pid 26592] AH02312: Fatal error initialising mod_ssl, exiting.
```

mikeashley:
On 2016-01-18 19:17, Allen Riddell wrote:
> On a fresh Jessie install I'm hitting a SSLStapling log. I think this
> might have been fixed somewhere else.
> 
> [Tue Jan 19 00:14:44.414523 2016] [ssl:emerg] [pid 26592] AH01958:
> SSLStapling: no stapling cache available
> [Tue Jan 19 00:14:44.414567 2016] [ssl:emerg] [pid 26592] AH02312:
> Fatal error initialising mod_ssl, exiting.
> 
> --
> Reply to this email directly or view it on GitHub [1].
> 
> 	*
> 
> Links:
> ------
> [1] 
> https://github.com/sovereign/sovereign/pull/451#issuecomment-172690364

Yes, I hit the same on the jessie branch.  I disabled SSLStapling for 
testing purposes.

ariddell:
FYI: LE supports DNS validation now. https://twitter.com/letsencrypt/status/689919523164721152

neuhaus:
That's cool, however DNS is not yet part of sovereign (it should be IMHO). I run djbdns/tinydns on my sovereign box.

ariddell:
DNS would be great to provide.

Here's the part of the ACME spec relevant to DNS verification: https://ietf-wg-acme.github.io/acme/#rfc.section.7.5

simontretter:
So as per the current master CONTRIBUTING.md, new modules are required to update the letsencrypt tasklist in the common role. Since it is still not merged, is this to be ignored?

mikeashley:
It should be ignored.  It should also be pulled since we're automatically enumerating the registered subdomains now.  I'll make sure that happens in the final PR.

mikeashley:
I looked at the Apache configuration changes discussed above to avoid stopping Apache during certificate generation and renewal.  It's not clear to me how to best do it.  Three possible ways:

1. Use Redirect, but this does not appear to be expressive enough.
2. Use mod_rewrite, but this isn't recommended.
3. Run a second web server, e.g., nginx, on port 80 and run Apache on 443.

A complication is that autoconfig is set up to serve HTTP and HTTPS due to ambiguity in the documentation (or so says the comments on autoconfig.conf).

Does Apache need to be restarted anyway after certificate renewal in order to load the new certs?  Some other services like dovecot do need to be bounced.

tynsh:
I think using mod_alias is the best way to do it. I am currently using an alias on nginx, but I've never tried it on apache. The [documentation](http://httpd.apache.org/docs/2.2/mod/mod_alias.html) says you just need to specify `Alias /.well-known/acme-challenge /var/www/letsencypt_challenge`

neuhaus:
Here is what I am doing on my server:
I want to have a redirect for all virtual hosts that redirects all requests from port 80 to https on port 443. At the same time, I want to add "www" if it's missing.
However, the requests for ACME should go through without being redirected.
So I put this block in every virtualhost listening on port 80:

```
<VirtualHost *:80>
        ServerName example.com
        Alias /.well-known/acme-challenge/ /var/www/challenges/
        <Directory /var/www/challenges>
                Order allow,deny
                Allow from all
        </Directory>

        RedirectMatch permanent "^(?!\/\.well-known\/acme-challenge\/)(.*)" "https://www.example.com$1"
        DocumentRoot "/var/www/example.com"
</VirtualHost>
```

Note that the domain name in question also appears in the `RedirectMatch` line.

mikeashley:
Ok.  I was trying to find a way to avoid a customization to every virtual host config, but I'll go ahead with your approach.  Thanks for the snippet.  Also thanks for the recommendation on apache restart.  I'll use that as well.

ariddell:
The idea of serving port 80 with nginx sounds appealing. It's consistent
with the way nginx gets used in other settings (e.g., as a proxy). We
could also move out all the redirect handling from Apache (from 80 to
443) which has caused some problems in the past. It's a nice separation
of concerns, IMHO.

(I'm also in favor of just switching to nginx although I recognize
that's a separate issue.)

mikeashley:
Any more input?  I am tempted to prototype the nginx path to see it looks cleaner.

@ariddell I am curious to know what problems the current redirect handling has caused in the past.

ariddell:
For example, #360 and #425. I wouldn't say these are necessarily
"problems" but I do think these are annoyances which would be
ameliorated by clearly separating the task of redirecting from 80
to 443 from the task of serving content on 443. It's hard to separate
the tasks any more than by having different programs do them. It's also
a separation which is consistent with the way nginx is often used.

ariddell:
I also love the idea (implicit, I realize) that port 80 is no longer for
serving traffic but really it's just a dedicated port for
"communicating" with an ACME service like Let's Encrypt.

mikeashley:
@neuhaus A think a bug with your recommended approach is that the initial certificate generation will fail, because the roles that put down the virtual host configurations have not run yet.  It will succeed on the second and subsequent runs of the playbook.

mikeashley:
@ariddell Ha!  #360 burned me on the blog role virtual host changes for LE as well.

neuhaus:
@mikeashley if no VirtualHost is configured in Apache it should use the default host. As long as the `Alias` is configured there, it will work.

@aridell I don't mind switching to nginx completely, but I'm not fond of using two different pieces of HTTP server software on one machine. It just wastes RAM (which is the most expensive resource on a rented server). Apache is perfectly capable of serving this redirect (as is nginx, of course).

We can try to get rid of all VirtualHost instances on port 80 and use mod_redirect on the default instance listening on port 80 to do the redirects and serve the ACME files if you guys prefer that.

That said, I'm not the one who is implementing this feature so if you love this idea, don't mind my babbling.

mikeashley:
Ok, I got enough feedback.  Thank you!

I am rescinding this PR.  I'll submit a new PR after the jessie catchup PR is merged and after I've finished a stab at not stopping Apache to generate or renew certificates.

ariddell:
Just to follow up on the question of providing one's e-mail address to Let's Encrypt:

[Let's Encrypt leaked thousands of email addresses, asks to not post lists](https://community.letsencrypt.org/t/email-address-disclosures-preliminary-report-june-11-2016/16867)

PR #452, opened 2015-12-27
Adding monit tunnel information to Readme

NickBusey:
It took me some time to figure out how to access the monit status page, hopefully this will save some others some time.


neuhaus:
Thanks. Personally I prefer using DynamicForward (SSH SOCKS proxy feature, but this will do)

PR #455, opened 2015-12-31
LetsEncrypt support take 2

mikeashley:
Happy New Year.  Here is a different approach to using LetsEncrypt to replace the wildcard certificate we currently use.  As before, this is a draft pull request to advance the discussion in issue #426.  It's incomplete but illustrates the approach.

In contrast to my first pull request #451, this generates only one certificate using Subject Alternative Names to list the subdomains that Sovereign roles may use.  Since we are not using a wildcard certificate, roles are always going to have their fingers in certificate generation.  Given that, I went for broke and listed subdomains explicitly.

Based on feedback from #451, I put the LetsEncrypt stuff in a separate task list but elected to keep it in the common role.  It just doesn't seem warranted to put it in a separate role.  Also, I added a cron script to renew the certificate periodically (once a month in this draft).

Feedback welcome.  I prefer this approach over what I did in #451.


vars/defaults.yml
```diff
@@ -13,6 +13,8 @@ main_user_shell: "/bin/bash"
 # encfs_password: (required)
 friendly_networks:
   - ""
+# letsencrypt_email: (required)
```
ariddell: I don't think email is required.

https://letsencrypt.readthedocs.org/en/latest/using.html#configuration-file

ariddell:
This looks really good. Thanks! I'll try to find some time to try it out.

vars/defaults.yml
mikeashley: Providing an email address is [highly encouraged](https://community.letsencrypt.org/t/register-unsafely-without-email/4342).

mikeashley:
I just pushed a rebase.  I realized that virtual host configuration of SSL can be abstracted just like it is now, so the changes to the blog, owncloud, and other similar roles aren't necessary.

I went ahead and made the necessary updates to the xmpp and ircbouncer roles.

vars/defaults.yml
ariddell: Maybe s/required/recommended/? I think it would be great if it worked without setting the email since the ACME protocol explicitly supports this.

mikeashley:
Moving discussion back to first pull request to consolidate and doing force pushes to update that branch.

Issue #417, opened 2015-09-30
Roundcube database not created


m0rth:
I am on a Debian 7 Wheezy 64bit KVM VM on Proxmox. I followed the exact steps from the Readme.
While running the Playbook I get following error:
```
NOTIFIED: [webmail | set roundcube password] ********************************** 
failed: [redacted.me] => {"changed": true, "cmd": ["sudo", "-u", "postgres", "psql", "-d", "roundcube", "-c", "ALTER USER roundcube with password 'redacted';"], "delta": "0:00:00.053525", "end": "2015-09-30 13:19:26.335245", "rc": 2, "start": "2015-09-30 13:19:26.281720", "warnings": []}
stderr: psql: FATAL:  database "roundcube" does not exist
```
When I run the playbook again, it doesn't fail at that point, but doesn't create the database.
I get the error "DATABASE ERROR: CONNECTION FAILED!", when I try to connect to Roundcube.

/var/log/postgresql/postgresql-9.1-main.log:
2015-09-30 13:19:26 UTC FATAL:  database "roundcube" does not exist

/var/log/roundcube/errors:
[30-Sep-2015 13:37:18 +0000]: DB Error: SQLSTATE[08006] [7] FATAL:  password authentication failed for user "roundcube"
FATAL:  password authentication failed for user "roundcube" (GET /)

vlow:
I have the same problem. Debian 7 Wheezy 64 Bit KVM VM on another hoster.

KyleOndy:
I am having this issue on 64bit Wheezy VMs with both Linode and Digital Ocean.

    failed: [redacted.io] => {"changed": true, "cmd": ["sudo", "-u", "postgres", "psql", "-d", "roundcube", "-c", "ALTER USER roundcube with password 'not_my_password';"], "delta": "0:00:00.096447", "end": "2015-12-09 16:42:03.521714", "rc": 2, "start": "2015-12-09 16:42:03.425267", "warnings": []}
    stderr: psql: FATAL:  database "roundcube" does not exist

I could not get the database working with `sudo dpkg-reconfigure roundcube-core` under the deploy user,

mikeashley:
This is addressed in the move to Debian 8.

PR #457, opened 2016-01-01
Roundcube 1.1.4

mikeashley:
Here is a proposal to move from 0.9.5 to 1.1.4 that stops using Debian packages that have been unreliable.

I aimed for idempotency and therefore leave the user with some post-deploy finalization work to do.  I don't really like this, but I think idempotency is more important, and this also sets up for easier upgrades in the future.  Is this tradeoff ok?


mikeashley:
Tests are passing for both wheezy and trusty.  The meaningful difference between the two is Apache 2.2 versus Apache 2.4.

neuhaus:
I'd rather wait for Xenial to upgrade Roundcube to 1.1.4 using their standard package. For jessie we can use the jessie-backports repository to get Roundcube 1.1.4.
Meanwhile, issue #417 needs to be fixed by someone with some spare time to look into it.

mikeashley:
I no longer trust the roundcube packages on debian.  The wheezy backport has been troublesome (see #417, and it's currently breaking in sovereign's test environment).  I suspect that the package is trying to do too much, roundcube simply isn't designed for packaging, or some combination of the two.  Upgrades through the package especially scare me at this point.

neuhaus:
If it turns out that the version 1.1.4 packages give us problems we can use this pull request. For now, let's stick with the packages. Thanks.

PR #458, opened 2016-01-01
Updated contribution guidelines

mikeashley:
I probably got this wrong, but I'm putting a stake in the ground based
on work I've done on the Roundcube module and adding Let's Encrypt
support to the common module.


neuhaus:
Looks good, thanks!

Issue #440, opened 2015-12-08
Properly support jessie


danmilon:
Per our discussion at #427, lets discuss here how we will properly support jessie (and/or upcoming releases).

Currently master is based on wheezy and trusty and we've got the [jessie](https://github.com/sovereign/sovereign/tree/jessie) branch, which contains wheezy-breaking changes.

I wanted to provision sovereign on my jessie server, so I picked up the jessie branch, but later found out that there are changes in master that I'd like to have, and then it started becoming a mess. I believe the idea is that the jessie branch merges master every now and then to be up to date.

I believe dropping support for one of the currently supported distros is not an option, right? They are all still deployed at scale I suppose.

Is it realistic to include jessie support in master through conditionals? For example #394 is not so trivial to solve. Someone who has a better picture of the overall thing, what do you think?

ariddell:
I'd vote for a sovereign "2.0" which exclusively supports jessie. The shift to systemd is a big one.

sread:
I think this is the right way to go as well, especially considering the security status of oldstable: one year after current stable (jessie) release (https://www.debian.org/security/faq#lifespan)

neuhaus:
We'll have the same situation on our hands in April 2016 with the release of Ubuntu 16.04 LTS ("Xenial").
I'd prefer to keep backward compatibility unless there is a good reason not to, at least for half a year or so until after we have started to support a new LTS/stable version to give people time to migrate and the new distribution time to settle.
So far I haven't seen anything that makes backward compatibility impossible. Did I overlook something?

ariddell:
The argument in favor of having a backwards incompatible release for
Jessie and Xenial (they'll be roughly compatible, right?) is that it's
easier to audit and maintain the ansible scripts when you don't have all
the special cases to worry about. Supporting Debian 7+8+14.04+16.06 is a
lot of diversity.

lukecyca:
This would be a good time to consider splitting off some roles into their own repos and using Ansible Galaxy to deploy them. Then each role's OS support can be declared independently, and the transition can happen in stages. Our more popular roles will likely get support for more distros/versions, and the less popular ones won't.

Using ansible galaxy would bring some other advantages: people could use our roles à la carte and they'd be listed in the galaxy directory.

ariddell:
That makes a lot of sense.

conorsch:
[...]

PR #462, opened 2016-01-10
Radicale CalDAV and CardDAV as an alternative to owncloud.

simontretter:
This role provides Radicale ~~with postgresql backend~~. The Radicale Project is a complete CalDAV (calendar) and CardDAV (contact) server solution. Check out http://radicale.org

People who don't need owncloud or fear the added attack surface to their server can use this for calendar and contact sync. Although both can run at the same time without conflict, a way to select either solution might be beneficial.


fengor:
Nice. Going to test the role next week but I fully approve the idea behind it.

roles/radicale/tasks/radicale.yml
```diff
@@ -0,0 +1,39 @@
...
+- name: Configure Radicale
+  template: src=config.j2 dest=/etc/radicale/config group=root owner=root
+  notify: restart radicale
+
+- name: Create radicale accounts
+  htpasswd: path=/etc/radicale/users name={{ item.name }} password={{ item.password }} crypt_scheme=ldap_sha1 create=yes state=present
+  with_items: radicale_accounts
+
+- name: Set firewall rules for radicale
+  ufw: rule=allow port={{ radicale_port }} proto=tcp
```
jplock: I think this line needs a ufw tag

simontretter: Thanks, I added the tag.

roles/radicale/handlers/main.yml
```diff
@@ -0,0 +1,2 @@
+- name: restart radicale
+  command: /etc/init.d/radicale restart
```
spk: Why not using ansible service module ?

roles/radicale/tasks/radicale.yml
```diff
@@ -0,0 +1,40 @@
+- name: Install radicale dependencies
+  apt: pkg={{ item }} state=present
+  with_items:
+    - apache2-utils
+  tags:
+    - dependencies
+
+- name: Install radicale
+  apt: pkg=radicale update_cache=yes
```
spk: This package is from official repository, I think update cache is already done in common role

neuhaus:
Nice! I don't have a huge amount of trust in OwnCloud either.

roles/radicale/handlers/main.yml
simontretter: Thanks for pointing that out. I changed it.

roles/radicale/tasks/radicale.yml
simontretter: Your are right.

simontretter:
@neuhaus the time I can devote to this is limited, but I will add documentation once the role is well tested. For now, I don't know of a nice way to migrate the data. Radicale and Owncloud technically don't collide, you can run them at the same time. The question is whether you want to tolerate the feature overlap in a standard installation. 
To develop, I am using vagrant / debian78. I am still waiting for progress on the letsencrypt role to start deploying on live servers.

Radicale works well on Android (davdroid), iOS and OS X. As for Thunderbird, I don't think it will work (then again, I have not found any Caldav / Carddav implementation that works well with Thunderbird).
Please have a look at the Radicale user documentation for specific clients:
http://radicale.org/user_documentation/

ariddell:
This is the first new role in awhile. Personally, I'd like to use this role rather than owncloud.

What about making this the first standalone galaxy role, sovereign-radicale. @lukecyca raised this issue recently here https://github.com/sovereign/sovereign/issues/440#issuecomment-162928760.

jplock:
LGTM!  I think we should merge this in as-is instead of using galaxy to keep everything together until we can decide later to break roles apart.

ariddell:
If this is going to be merged into master and jessie then there needs to be another PR for jessie.

jplock:
One thing I just noticed, what version of radicale is available through the package?  They also publish versions via pip and the latest release is 1.1 which is not backward compatible.  I'm thinking we should try to source the package from pip and lock the release version.

neuhaus:
Wheezy currently has a package of radicale version 0.6.3-1
Trusty has radicale 0.8
Jessie has radicale 0.9
Xenial will have radicale 1.1.1

Does anyone know more about the compatibility problems that we may be facing?
If there is a problem getting these different versions to work we should consider doing the [PyPi installation of v1.1.x](https://pypi.python.org/pypi/Radicale/)

simontretter:
@neuhaus Here you go. This is untested as I am traveling. 

I very much dislike mixing different package management software and having to package myself. In this instance, I see no other option.

neuhaus:
Thanks! i hope I can test this a bit this weekend.

neuhaus:
@simontretter The patch contains the file `roles/radicale/files/etc_logrotate.d_radicale` which is not used. It's a copy of `roles/radicale/files/etc_logrotate_radicale` which is the one that gets copied to `/etc/logrotate.d/radicale`

simontretter:
@neuhaus I removed it.

mikeashley:
I am interested in this given the changes with the ownCloud organization.  @simontretter Have you been using this role?  Are there any changes you recommend?  I really appreciate the work you've put into this.

For those looking for an alternative to ownCloud, is there a recommendation for how to support file syncing?  That is mostly what I use ownCloud for today.

simontretter:
@mikeashley There still is some work needed. First problem as mentioned is that to have consistency through all supported linux flavours, we would need to use the pip version. I would want to avoid that (or at least wait for a proposal on how to update pip installed packages). 

Any suggestions on how to handle this? I would put some more work in next week to test the recent radicale version and then give you a more qualified answer.

mikeashley:
I am not familiar with how Python package management works so can't offer any guidance on how to best approach it.  For what it's worth, the jessie branch supports just Debian 8 today, but I anticipate that somebody will add support for Xenial after we switch it over to master.

simontretter:
@mikeashley here is an update. This installs radicale via apt on Ubuntu >=16. Debian and older Ubuntu versions get the pip variant. I had to switch from the database backend to the filesystem one because that seems to be the only reliable one. Does /decrypted/radicale seem like a good spot to put the data? I also integrated radicale with apache. This means radicale now goes through https://dav.mydomain.com on port 443 and apache handles TLS. 

Now what's left to do is to properly implement the apache authentication to replace the radicale one.

http://radicale.org/user_documentation/#idauthentication

mikeashley:
Thanks for the update.  Have you updated your branch to run under Jessie?  There is a merge conflict suggesting you are still basing on the old master (now the wheezy branch).

simonplend:
I don't want to take this thread too far off topic, but as someone planning to get sovereign up and running, I was wondering what is causing people's mistrust of ownCloud. @neuhaus and @mikeashley, can you shed some light for me? Thanks!

cimm:
I rebased this branch on the master branch and ran into some minor glitches while setting up a new Sovereign system today.

1. The `python3-passlib` seems no longer be available in Debian 8.6, I simply removed all `python3` packages from the [radicale.yml](https://github.com/simontretter/sovereign/blob/13ec85d65f92e1f9c0c7971efefd953780f47240/roles/radicale/tasks/radicale.yml#L4-L10) file, it seems to work without them.
2. It references the `{{ selfoss_domain }}` variable the [etc_apache2_sites-available_radicale.j2](https://github.com/simontretter/sovereign/blob/13ec85d65f92e1f9c0c7971efefd953780f47240/roles/radicale/templates/etc_apache2_sites-available_radicale.j2) file which does not work when you don't install selfoss. Why doesn't it simply depend on the `{{ domain }}` variable? The error log also points to selfoss, a copy & paste oversight?
3. Minor one: old style variables are depreciated in newer Ansible versions so [`radicale_accounts`](https://github.com/sovereign/sovereign/pull/462/files#diff-b09d9d7eea6c54e7c1f737d48e0cc2a3R55) becomes `'{{ radicale_accounts }}'`.
4. The SSL config in the [etc_apache2_sites-available_radicale.j2 ](https://github.com/sovereign/sovereign/pull/462/files#diff-ba2fec315fa8bfa74d861d521237cd18R10) template can't be loaded by Apache.

Thanks for this branch, really prefer a simple CardDAV and WebDAV solution over the full fledged OwnCloud! The only thing that's still missing in Sovereign. :wink:

simontretter:
I'm closing this due to a lack of traction. At this point, I thing more effort should go into documentation and maintenance of existing features than adding new ones.

PR #466, opened 2016-01-15
Become

fengor:
seperate pull request for sudo -> become change.

Upped requirements.txt to ansible 1.9.4

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-ansible==1.6.6
+ansible==1.9.4
 
--- a/roles/common/templates/etc_ssh_ssh_config.j2
+++ b/roles/common/templates/etc_ssh_ssh_config.j2
@@ -6,4 +6,5 @@
     GSSAPIDelegateCredentials no
     MACs {{ macs }}
     PasswordAuthentication no
+    UseRoaming no
 
--- a/roles/common/templates/etc_ssh_ssh_config.j2
+++ b/roles/common/templates/etc_ssh_ssh_config.j2
@@ -6,5 +6,4 @@
     GSSAPIDelegateCredentials no
     MACs {{ macs }}
     PasswordAuthentication no
-    UseRoaming no
 
--- a/site.yml
+++ b/site.yml
@@ -2,7 +2,7 @@
 # This is the top-level playbook that defines our entire infrastructure.
 
 - hosts: all
-  user: deploy
+  become_user: deploy
   become: True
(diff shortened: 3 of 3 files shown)


requirements.txt
```diff
@@ -1,2 +1,2 @@
-ansible==1.6.6
+ansible==1.9.4
```
neuhaus: why 1.9.4?

fengor:
because become is definately available in 1.9.4

ariddell:
The version string should be ``ansible>=1.9.4,<2.0``, IMHO. Could this and #487 be reconciled somehow? The PRs are not very clear. From the PR text and commit messages I'm not sure what the problem is and/or why the commits solve the problem.

ariddell:
What about a single PR that deals with upgrading to Ansible to 2.0+ and deals with the ansible-lint issue in #487. It should have a title like "Upgrade Ansible to 2.0". I'm willing to do this if nobody else gets around to it.

neuhaus:
We don't know what version of ansible the users have. Ansible (often) runs on a different machine than the one that is being configured. If it's a Ubuntu trusty box they probaby have Ansible 1.5.4. Debian jessie ships with Ansible 1.7.2

carljm:
There is no reason to rely on a system-installed Andible version; a project like sovereign needs to be able to pin itself to a specific Ansible version and rely on that. The instructions should include creating a Python virtualenv and running `pip install -r requirements.txt` within it, so we always have exactly the specified Ansible version available.

ariddell:
I agree. Let's just decide on a version to support and have the
requirements.txt file reflect this. I believe Ansible is using semantic
versioning so >=2.0,<2.1 sounds good to me.

nodiscc:
For what it's worth many users don't want to depend on pip installs for production/trusted machines. I prefer sticking to the ansible version provided by my package manager. As your current target is Jessie, I think you should also expect ansible to be at the version provided in Jessie (even if the admin/deployement machine is not always using the same OS/version).

Just my 2 cents.

ariddell:
IIRC, we already required a version higher than the version installed by default with wheezy/trusty.

Remember, we're talking about the local copy of ansible. Having a virtualenv or using ``pip install --user`` is no big deal.

ariddell:
Since master is doing fine could we apply this just to jessie? We could upgrade ansible to 2.0 and ansible-lint as well.

ariddell:
Fixed with #508.

PR #473, opened 2016-01-28
Add SSL stapling cache to correct configuration

mikeashley:
The cache configuration must be global, i.e., not per virtual host.  As
a result, this change moves most of the SSL configuration to a global
configuration and leaves enabling the SSL engine to each virtual host
that wants to use it.

### DIFF
--- a/roles/blog/templates/etc_apache2_sites-available_blog.j2
+++ b/roles/blog/templates/etc_apache2_sites-available_blog.j2
@@ -9,8 +9,7 @@
 <VirtualHost *:443>
     ServerName {{ domain }}
     ServerAlias www.{{ domain }}
-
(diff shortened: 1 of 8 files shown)


ariddell:
I think master got a patch on this. Let me finish up cherry-picking stuff (ten minutes) and see if it includes the fixes.

mikeashley:
Looks good.  I like moving the SSL configuration to global instead of using an Include in the virtual hosts, but it's a style thing.  I may propose this again, but I'm eager to see the catchup PR merged so we can finalize the LE PR and get it on jessie.

mikeashley:
Rescinding this and will resubmit after the jessie catchup PR is merged.

PR #477, opened 2016-02-01
Fix condition for rspamd Ubuntu repo installation.

carljm:


carljm:
I'm also curious -- why the `download.suse.com` repos for rspamd, instead of the `rspamd.com` repos recommended at https://rspamd.com/downloads.html ? And why using repos labeled for Ubuntu 14.10 for 14.04?

neuhaus:
Good questions. Looks like these are issues that should be fixed.

carljm:
Shall I add the fixes for those two issues to this PR as well, or file separate PRs?

neuhaus:
Yes please

carljm:
Replaced by #486.

PR #446, opened 2015-12-13
Pin Selfoss version to an actual release

jplock:
Cherry pick this commit from the `jessie` branch

PR #486, opened 2016-02-11
Update rspamd repository to the official one.

carljm:
This replaces #477.

PR #487, opened 2016-02-11
Fix travis build by fixing ansible-lint version

spk:
cf https://github.com/sovereign/sovereign/pull/486


ariddell:
I'll merge this but what exactly does it fix?

carljm:
Should fix the Ansible lint errors on #486, which are unrelated to that PR.

spk:
Yes we have to use last ansible version for syntax check and a fixed version for ansible-lint.
I don't have better solution right now.

ariddell:
Sorry, I'm confused. Why can't the requirements.txt file be used? Shouldn't the travis.yml file just have a line pip install -r requirements.txt?

spk:
No I tried in this commit https://github.com/spk/sovereign/commit/3daf7cf56bb1d84d9efbd78e743569748251de48
but ansible-lint failed completely due to too old ansible version.
cf https://travis-ci.org/sovereign/sovereign/builds/108639408#L443

ariddell:
So the problem is with ansible. Why are we pinning to 1.6.6 anyway?

Can you go back to spk@3daf7cf and figure out a version of ansible that works?

If you're feeling up to it, the ansible-lint requirement probably should go in test-requirements.txt. I believe that's the convention.

spk:
I cannot upgrade ansible, this is a known issue, cf https://github.com/sovereign/sovereign/pull/466
I did not know about test-requirements.txt, I will do that.

.travis.yml
```diff
@@ -1,7 +1,11 @@
 language: python
 python: "2.7"
+cache:
+  directories:
+    - $HOME/.cache/pip
 install:
-  - pip install ansible ansible-lint
+  - pip install ansible
```
ariddell: this should be ``pip install -r requirements.txt`` (to get ansible)

requirements.txt
```diff
@@ -1,2 +1 @@
 ansible==1.6.6
```
ariddell: What happens if this is something like
``ansible>=1.9.4,<2.0`` instead of a fixed version?

test-requirements.txt
```diff
@@ -0,0 +1 @@
+ansible-lint==2.3.6
```
ariddell: Is this the only version of ansible-lint that works or can we use a version range like
``ansible-lint>=2.3.6,<3.0``

.travis.yml
spk: Should I rebase on `jessie` branch ? because on `master` ansible version `1.6` is too old ?

test-requirements.txt
spk: Only ansible-lint `2.3.6` is working after there will be sudo/become errors that are not merged yet.

ariddell: This PR is confusing to me. What problem is being solved here? A more
descriptive PR text would be helpful. Travis is currently building
successfully on master.

spk:
Title updated. Currently ansible is fixed to `1.6.6`, this version does not work with ansible-lint.
And there is still `sudo` on the codebase that cause errors so fixed ansible-lint to 2.3.6

ariddell:
OK. I think I get this now. Sorry for being so slow.

Can you squash/rebase all those commits into a single commit (since it's a simple change)?

p.s. maybe then we can resolve #446 by upgrading both ansible and ansible-lint to the current version (maybe just on jessie)?

spk:
No problem @ariddell 
This is done and I also added a todo on `.travis.yml`

PR #488, opened 2016-02-13
Fix version of tomcat

mikeashley:
Both jessie and wily use the standard solr package, which expects tomcat7.  This PR switches from tomcat8 to tomcat7.  This prevents tomcat7 and tomcat8 from both being installed and thus fighting over port 8080 and generating monitoring alerts.


ariddell:
LGTM

chrisbanm:
So, I'm on Debian 9.12 and solr wants tomcat8 but all of the scripts reference tomcat7. I also noticed that the travis build has failed for PR #488. What should I do? Change all the scripts to use tomcat8 but simple replacement of the string tomcat7 for tomcat8?

Issue #474, opened 2016-01-28
ircbouncer role fails to run


mikeashley:
The package installs the znc client but does not set it up as a service.  Thus the role fails to run because the user 'znc' does not exist.

There is a bug/enhancement report in to the package maintainer to install as a service.  We could either wait and see if that gets done, or we could go back to the old way of installing.

mikeashley:
Fixed with PR #490.

PR #490, opened 2016-02-13
Fix znc configuration

mikeashley:
The znc package installs the client but does not set it up as a service.  This patch restores the service configuration that was done on wheezy/trusty.

This addresses Issue #474.


ariddell:
Should we try to use systemd instead of init in jessie? Surely there's a systemd service file we can copy from somewhere? I find the ``.service`` files far more readable at this point.

mikeashley:
Good suggestion.

PR #493, opened 2016-02-13
Remove references to Trusty and Wheezy

mikeashley:
Make a clean distinction between Debian 7 and Debian 8.  Anticipate the next Ubuntu LTS release (Xenial) that is planned for support.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@
 
 Make sure you agree with the license (GPLv3). See [LICENSE.md](./LICENSE.md) for details.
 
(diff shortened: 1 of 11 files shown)


carljm:
This conflicts with #486. 

I hope having things actually working on xenial isn't going to be a blocker for the Jessie branch, since xenial is in pretty early stages still. E.g. The official rspamd apt repo (which #486 switches to) doesn't even have xenial packages yet.

mikeashley:
This also conflicts with PR #489.  I recommend delaying this PR.  I'll fix it up after the two conflicting PRs are resovled.

Issue #495, opened 2016-02-14
Support roundcube


mikeashley:
Roundcube support needs to be reinstated from Debian backports and tested.  We've had nagging installation problems (#417 is the latest) under wheezy.  Is the situation any better under jessie, or do we need to consider a different install method?

neuhaus:
For the record, the [version](https://packages.debian.org/en/jessie-backports/roundcube) available via jessie-backports (1.1.4) is the same that Ubuntu Xenial [will ship with](https://launchpad.net/ubuntu/xenial/+source/roundcube).

neuffer:
Current roundcube in Debian testing is1.2.0+dfsg.1-1

mikeashley:
Addressed with PR #590.

PR #498, opened 2016-02-18
Remove references to Trusty and Wheezy (jessie)

mikeashley:
Make a clean distinction between Debian 7 and Debian 8.  Anticipate the next Ubuntu LTS release (Xenial) that is planned for support.  This drops roundcube, but we have Issue #495 noting that we need to restore it.

This deploys with no errors in the Vagrant test environment.  I'm expecting this should help streamline work to migrate over to Jessie and get remaining problems straightened out before adding Xenial support.


ariddell:
Thanks for this.

Perhaps there should be a separate commit removing roundcube (all of it, including the role) pending a real solution. The code for the role is still in the git tree. If the roundcube removal is its own commit it could also be trivially reverted.

mikeashley:
That works for me.  I was a little reluctant to drop it, but Roundcube has gone from 0.9 to 1.2, so I think the whole role needs pretty careful review before reinstatement.

mikeashley:
Missed a couple of references to Ubuntu in the common role.  Corrected.

ariddell:
Thanks for this. We're almost there!

neuhaus:
So the reason you've removed roundcube completely is because it no longer ships with Debian?
It looks like Ubuntu Xenial will ship with Roundcube 1.1.4... I think it's a highly useful role, we offer no other way to configure SIEVE mail filtering rules on the server!

neuhaus:
Roundcube 1.1.4 [is available from jessie-backports](https://packages.debian.org/en/jessie-backports/roundcube)

Issue #475, opened 2016-01-28
openvpn role fails to run


mikeashley:
The dnsmasq service fails to restart.

● dnsmasq.service - dnsmasq - A lightweight DHCP and caching DNS server
   Loaded: loaded (/lib/systemd/system/dnsmasq.service; enabled)
  Drop-In: /run/systemd/generator/dnsmasq.service.d
           └─50-dnsmasq-$named.conf, 50-insserv.conf-$named.conf
   Active: failed (Result: exit-code) since Thu 2016-01-28 01:38:16 UTC; 52s ago
  Process: 18487 ExecStop=/etc/init.d/dnsmasq systemd-stop-resolvconf (code=exited, status=0/SUCCESS)
  Process: 18518 ExecStart=/etc/init.d/dnsmasq systemd-exec (code=exited, status=2)
  Process: 18517 ExecStartPre=/usr/sbin/dnsmasq --test (code=exited, status=0/SUCCESS)
 Main PID: 17071 (code=exited, status=0/SUCCESS)

Jan 28 01:38:16 li879-96 dnsmasq[18517]: dnsmasq: syntax check OK.
Jan 28 01:38:16 li879-96 dnsmasq[18518]: dnsmasq: failed to create listening socket for 10.8.0.1: Cannot assign requested address
Jan 28 01:38:16 li879-96 dnsmasq[18518]: failed to create listening socket for 10.8.0.1: Cannot assign requested address
Jan 28 01:38:16 li879-96 dnsmasq[18518]: FAILED to start up
Jan 28 01:38:16 li879-96 systemd[1]: dnsmasq.service: control process exited, code=exited status=2
Jan 28 01:38:16 li879-96 systemd[1]: Failed to start dnsmasq - A lightweight DHCP and caching DNS server.
Jan 28 01:38:16 li879-96 systemd[1]: Unit dnsmasq.service entered failed state.

mikeashley:
I investigated this.  It looks like ansible is reporting that the openvpn server is started after the handler runs to bounce it, but it's not, and thus dnsmasq has no interface to bind against when it tries to start.  Logging into the server and starting openvpn followed by starting dnsmasq works fine.

Running ansible in verbose mode (`vvvv`) wasn't very helpful.  Root cause still needs to be identified.

mikeashley:
Issue #401 is related.

mikeashley:
Fixed with PR #500.

PR #500, opened 2016-02-19
Fix systemd configuration of OpenVPN server

mikeashley:
This addresses Issue #475.  Runs under Vagrant.

The server was not starting.  As a result, the dnsmasq service failed to
start, and the playbook thus failed to run when using the vpn role.  This
patch corrects the configuration per instructions from
https://help.ubuntu.com/community/OpenVPN.

OpenVPN PAM configuration moved up to reduce server bouncing as the
playbook runs.  The dependency on service (re)starts between openvpn and
dnsmasq works but feels brittle.

### DIFF
--- a/roles/vpn/handlers/main.yml
+++ b/roles/vpn/handlers/main.yml
@@ -2,4 +2,4 @@
   service: name=dnsmasq state=restarted
 
 - name: restart openvpn
-  service: name=openvpn state=restarted
+  service: name=openvpn@server state=restarted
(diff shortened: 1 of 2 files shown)


mikeashley:
Updated commit message, no change to patch.

ariddell:
Thanks. I remember working on this briefly and trying to get the dnsmasq / iptables setup working in a sane manner.

Frankly, if it feels too brittle I wouldn't object to removing OpenVPN and having an issue to fix it properly. I think openvpn is one of the roles which needs to change the most with systemd

Issue #502, opened 2016-02-21
Run opendmarc reporting with postgres as the back end


mikeashley:
The opendmarc package uses mysql as the back end.  If the installation could be modified to use postgres, this would allow sovereign deploys to avoid installing mysql.  There was a brief discussion around this in the comments on PR #397.

neuhaus:
This would also fix issue #497 if it were done on trusty (it's not an issue for xenial).

ariddell:
What needs to be done to make this happen. It would be a positive development if only postgres were required.

mikeashley:
At a minimum the SQL to initialize the database needs to be converted from MySQL to PostgreSQL, and the package needs to be configured to use posgres as the backend.  Nominally it supports this, since the backend support looks abstracted, but I don't recall seeing anything to suggest that this had ever been tested or is supported.

sorki:
I've created an experimental fork of opendmarc with PostgreSQL support (only)

https://github.com/nestihacky/opendmarc

Commit: https://github.com/nestihacky/opendmarc/commit/0241bf1999706b197ccb021da75a921bbc38233f 

I also have packages for Fedora/Centos if you would like to test it: https://copr.fedorainfracloud.org/coprs/rmarko/opendmarc/

I'm only running this for few days in production. It still needs a number of IF clauses to handle both MySQL and PostgreSQL, bit more testing and then I'll try to upstream this.

Btw we now have usable mailserver deployment scripts based on sovering mailserver that targets Centos 7 - https://github.com/nestihacky/ansible-role-mailserver

neuffer:
what is the status on this issue? 
Would it be possible to use opendmarc with postgresql via the libopendbx1-pgsql backend?

mikeashley:
Addressed in PR #720.

Issue #491, opened 2016-02-13
OpenDMARC fails to install (jessie)


mikeashley:
The database import fails to run.  There is a syntax error in the import file, and after that is fixed, there is type error.  It appears the import file was written for a database other than postgres.

It looks like this broke at commit 2352d2d67e591cb1ee841c0782ea7ce05b36e355.  Speculation: the aim of the commit was to remove the last(?) reference to mysql.

mikeashley:
I would like @al3x to weigh in on intent.  Some brief googling fails to turn up how to configure this to use postgres, and the man pages are all about mysql for report generation.  Seems like the two options are 1) drop opendmarc, and 2) revert to using mysql and install it.

al3x:
Yes, I was in the process of removing MySQL. It seemed to bother users that they were running two databases (understandably).

mikeashley:
I found Issue #397 where this has come up before.  I recommend reverting to MySQL so that it works and that we can keep moving towards a stable jessie branch.  If somebody feels strongly about squashing MySQL from the system, they can tackle making it work with Postgres separately.

PR #503, opened 2016-02-21
Fix opendmarc installation

mikeashley:
This addresses Issue #491.  Running under postgres is a good goal, but let's get the jessie branch stable.  Switching to postegres will take some feasibility work.  I opened Issue #502 to track it.


ariddell:
Shame about putting back mysql but I suppose this makes sense.

PR #504, opened 2016-02-21
Let's Encrypt support

mikeashley:
Here is the proposal for LE support again.  I would still like to avoid bouncing Apache in order to generate and renew certificates, but I would also like to get this into testing sooner rather than later.  If this PR is merged onto jessie, I'll add a ticket to address the bounce.


ariddell:
I started reviewing this. I'll do more today or tomorrow.

ariddell:
Got distracted this week. If someone else wants to look through the commits please go ahead. If not, I'll try to look at it soon.

ariddell:
I've reviewed the code today. It all looks good.

mikeashley:
Indeed, an earlier version used a variable in defaults.yml to say what domains the certificate should include.  I abandoned it, though.  The Let's Encrypt certificate server uses DNS to verify sites for certificate generation, so DNS must be configured before the playbook is run regardless, and the extra variable in defaults.yml introduced a possibility for the playbook configuration to be out of sync with the DNS configuration.

mikeashley:
Good suggestion on the script name.

ariddell:
Great. Thanks!

roles/common/files/etc_cron-monthly_letsencrypt-renew
```diff
@@ -0,0 +1,21 @@
...
+# Given a certificate file returns "domain1,domain2"
+# https://community.letsencrypt.org/t/help-me-understand-renewal-config/7115
+function getDomains() {
+        openssl x509 -text -in "$1" |
+        grep -A1 "Subject Alternative Name:" | tail -n1 |
+        tr -d ' ' | tr -d 'DNS:'
+}
+
+service apache2 stop
+for c in `ls /etc/letsencrypt/live`; do
```
nodiscc: `ls` output cannot be parsed reliably, one should use `find` instead.

ariddell:
Looks like Google's Bash style recommends
```
while read f; do
  echo "file=${f}"
done < <(ls -l /tmp)
```
https://google.github.io/styleguide/shell.xml?showone=Variable_expansion#Variable_expansion

nodiscc:
@ariddell a reputable source is at http://mywiki.wooledge.org/ParsingLs but I guess keeping `ls` is fine as long as it's not demonstrably broken in this case.

ariddell:
I think it is definitely something worth changing either to use ``find`` or to use the Google method. @nodiscc do you want to submit a PR.

nodiscc:
@ariddell why not? Using find + the `$(command)` notation instead of backticks would be cleaner. I would also use `set -o errexit` so that the script exits with non-zero return code if something goes wrong. This way the ansible task would fail and the user would be notified. Is this ok?

ariddell:
These are all important changes! I agree about errexit and avoiding backticks. Did I mistype something? (This PR has already been merged, so there needs to be a new PR with the fixes)

roles/common/files/etc_cron-monthly_letsencrypt-renew
neuhaus: :+1:

ariddell:
Finally tested this live. Everything appears to be working. Thanks again @mikeashley

carljm:
I should mention that I've got it running live as well and haven't noticed any issues. So far just the common and mailserver roles (haven't done anything web-related yet), but it's working fine for SMTP and IMAP.

PR #505, opened 2016-02-28
Update ownCloud role for ownCloud 8.2

mikeashley:
- Remove unnecessary dependencies
- Clean up postgresql server setup
- Install from official package repository
- Expect package installer to enable modules that are needed
- Update virtual host config with owncloud 8.2 configuration
- Update post-deploy instructions


mikeashley:
This update deploys in the vagrant test environment and passes some basic smoke testing including file upload/download and generating expiring links.

PR #506, opened 2016-03-04
Restore webmail role for jessie

mikeashley:
Address Issue #495.  This has been lightly tested on a sandbox deployment.


mikeashley:
I used debconf to preconfigure the package manager to use postgres for the backend.  This might help with the install problems we've seen.

mikeashley:
Using debconf doesn't seem to be working on master.  I'd like to see if I can find a solution that works for both branches.  I'm going to close this PR in order to work on it a bit more.

PR #507, opened 2016-03-05
Reinstate webmail role

mikeashley:
Address issue #495.  This has been lightly tested on a sandbox server but should get more testing.

I like using debconf to get the package configured correctly and avoid whacking configuration files into place.  I tried to make this work on mater, but I could not seem to get debconf changes to be used by the roundcube package.  I am probably doing something wrong.  It works fine here on the jessie branch though.


mikeashley:
Any holdup on reviews of this?  Anything I can do to facilitate?  The failing checks are because this was submitted before the travis fix.

ariddell:
I don't know anything about webmail and I've never used it. Perhaps someone else can review this?

neuhaus:
I can have a look at this

roles/webmail/DESIGN.md
```diff
@@ -0,0 +1,19 @@
+# Design Description for Webmail Role
+
+## Scope
+
+The goal of this role is twofold: 1) provide an interface for configuring dovecot sieves, and 2) provide a web-based front-end for email stored on the server for users who want access "anywhere" without a mobile device.
+
+## Approach
+
+The Roundcube package is used for this role.  It's provided in backports on Debian and included in Xenial.  Plugins are used for optional two-factor authentication for additional security and for CardDAV support.
```
neuhaus: SIEVE support is also provided by a plugin

mikeashley:
This is not ready to go.  Looks like there is a bug in the package configuration when running from a clean install.

mikeashley:
Fixed with latest commit.

mikeashley:
This is not ready to go, still.  `debconf` is fussing about missing questions in some setups.

mikeashley:
I'm not able to reproduce the debconf problems I encountered while upgrading from Wheezy to Jessie when doing a clean install.  The debconf tasks in `roundcube.yml` run fine.  They update the configuration database correctly.

I don't think it's worth the effort to root cause and find a fix for a problem that only occurs on upgrades.  I suggest instead we document the workaround on the wiki.

@neuhaus Any opinions on this?  I think this PR is still our best proposal for how to use the packaged distribution of roundcube correctly.

neuhaus:
So you tried this both with a fresh jessie install and an installation of wheezy upgraded to jessie?

mikeashley:
I've tested with fresh Jessie installs with Vagrant and a test instance at linode.  I tried to reproduce the problem and investigated within Vagrant.  Indeed, the next step would be to try with a Wheezy install upgraded to Jessie.

mikeashley:
I tried to reproduce with a Debian 7.9 VM upgraded to Jessie and was not successful.  It's possible the problem was unique to my production server and the way I executed the upgrade.

mikeashley:
Are we ready to merge this PR?  It's an important one to return feature parity with master.

mikeashley:
This will need updated after #526 is merged.

roles/webmail/templates/etc_roundcube_config.inc.php.j2
```diff
@@ -0,0 +1,1144 @@
...
+// %n - hostname ($_SERVER['SERVER_NAME'])
+// %t - hostname without the first part
+// %d - domain (http hostname $_SERVER['HTTP_HOST'] without the first part)
+// %z - IMAP domain (IMAP hostname without the first part)
+// For example %n = mail.domain.tld, %t = domain.tld
+$config['smtp_server'] = 'ssl://{{ mail_server_hostname }}';
+
+// SMTP port (default is 25; use 587 for STARTTLS or 465 for the
+// deprecated SSL over SMTP (aka SMTPS))
+$config['smtp_port'] = 465;
```
neuhaus: change to submission port if pull request #560 is merged

neuhaus:
Needs a small update if pull request #560 is merged

aminb:
I was testing this branch and I got this error:

```
TASK: [webmail | Configure carddav database] ********************************** 
failed: [xxx.xxx.xxx.xxx] => {"changed": true, "cmd": "PGPASSWORD='redacted' psql -h localhost -d roundcube -U roundcube -f /usr/share/roundcube/plugins/carddav/dbinit/postgres.sql", "delta": "0:00:00.075856", "end": "2016-06-04 04:28:49.057581", "rc": 2, "start": "2016-06-04 04:28:48.981725", "warnings": []}
stderr: psql: FATAL:  password authentication failed for user "roundcube"
FATAL:  password authentication failed for user "roundcube"

FATAL: all hosts have already failed -- aborting
```

I think it's the same issue as with #417, and I remember having that on wheezy too.

Does any else still have this problem on jessie?

Ultimately, I had to comment that task and remove carddav from the list of plugins for the playbook to continue.

mikeashley:
What happened is that roundcube's debconf selections were set incorrectly and you were left with a broken roundcube install.  I have not had the chance to root cause this and find the correct fix.

The workaround is to remove and purge the roundcube packages, and then reinstall the packages manually.  Use postgres as the backend.  Once you've done this, the playbook should run without error.

aminb:
@mikeashley Thank you! Your workaround helped resolve the problem, and all is fine.

I was looking at the diff between the file created by manual install and the one generated by sovereign's templates, and I couldn't pin down the problem. But at least I have a working setup, so thank you.

mikeashley:
Just an update.  I've chased this into the noninteractive install of the roundcube-core package.  I may be doing something wrong with options, but it sure feels like a bug in there.  I've filed a [bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=829485) with the maintainers.

mikeashley:
There has been no response from the package maintainers on my bug report.  We have three choices:

1. Continue to wait for a response.
2. Back off and install roundcube using mysql as the back end.
3. Install the package from source.

I am against 3 since in principal we would be giving up easy security and version updates.  I don't want to burden every user of sovereign with having to manually upgrade.

I am leaning towards 2.  We already have to use mysql for opendmarc, and while I dislike adding another dependency on mysql, I think the urgency to resolve this issue outweighs it.

Any opinions?

neuhaus:
What do you think about using sqlite? It would mean no extra server process and the load on the database is likely to be low.

mikeashley:
I suggested mysql since that is what Roundcube defaults to, and I think using the default DB backend is going to be the only way to install the current package noninteractively.

There is an Option 4, which is to fix the roundcube package installer and submit it upstream, but that would be new territory for me and definitely take longer than pursuing Option 2.

All that said, I'm a big fan of sqlite and do think it's an appropriate backend.

neuhaus:
I think we should try to get rid of a second database (which is MySQL) if possible, so adding a new dependency is not the way to go.
For dedicated servers, RAM is by far the most expensive resource. Having two database servers running is a relatively big memory hog.
Sqlite is among Roundcube's [currently supported database providers](https://github.com/roundcube/roundcubemail/wiki/Configuration).

mikeashley:
I am closing this PR, backing off the approach of using backports.  I'll submit another PR for installing from source.

PR #508, opened 2016-03-07
Update ansible in jessie

ariddell:
Use ansible>1.9,<2 on jessie.

Resolves #466 on jessie and contains the jessie version of  #487 (applied to master)

(I attempted to use Ansible 2.0 but a variety of new errors popped up.)


mikeashley:
Ran this with Vagrant and reviewed the diffs.  Looks good.

PR #511, opened 2016-03-07
DRY in openvpn role and allow other rc.local tasks

Yannik:


mikeashley:
Was this change tested?

Yannik:
I personally do use these tasks in my own openvpn role.

mikeashley:
Great.  Thanks for this.

ariddell:
We need one of these for jessie too, I think.

PR #513, opened 2016-03-07
Remove stray + prefix in sudoers file.

carljm:
This `+` prefix in the `/etc/sudoers.d/sudoers` file created by Sovereign prevented my main user from actually being able to sudo. Per the sudoers man page, a `+` prefix indicates a "netgroup" -- but this is clearly a user name, not a netgroup, so I'm not sure why the `+` is there. Removing it allowed my user to sudo.


carljm:
Wow, the initial `+` and `-` in the diff lines make this a confusing change to review :-) I wonder if diff/patch issues explain how this `+` snuck in there to begin with.

ariddell:
You can do a git bisect

carljm:
I did; the `+` appears to have been there since that file was first added.

PR #516, opened 2016-03-08
Make OpenDMARC cron job log everything instead of mailing some output to root.

carljm:
With current code in jessie, I get this daily email from cron to `root@mydomain`:

```
opendmarc-import: started at Tue Mar  8 02:00:02 2016
opendmarc-import: connected to database
opendmarc-import: updating at line 19
opendmarc-import: updating at line 36
opendmarc-import: terminating at Tue Mar  8 02:00:02 2016
opendmarc-reports: started at Tue Mar  8 02:00:02 2016
opendmarc-reports: selected 1 domain(s)
opendmarc-reports: terminating at Tue Mar  8 02:00:02 2016
opendmarc-expire: started at Tue Mar  8 02:00:02 2016
opendmarc-expire: connected to database
opendmarc-expire: expiring messages older than 180 day(s)
opendmarc-expire: expiring signatures on expired messages (id < 1)
opendmarc-expire: expiring request data older than 180 days
opendmarc-expire: terminating at Tue Mar  8 02:00:02 2016
```

It looks like the opendmarc report cron job is supposed to send its output to `/var/log/opendmarc_report.log`, but it's currently redirecting only stdout, not stderr, so the stderr output gets mailed to me.

Is this intentional? Is this a report that I should be carefully reading each day?

Assuming it's not intentional, this PR redirects stderr as well, so all output goes to the log file and nothing gets mailed to root.

### DIFF
--- a/roles/mailserver/tasks/opendmarc.yml
+++ b/roles/mailserver/tasks/opendmarc.yml
@@ -38,5 +38,4 @@
   file: path=/var/run/opendmarc/opendmarc.dat state=touch owner=opendmarc group=opendmarc
 
 - name: Activate OpenDMARC report cronjob
-  cron: name="OpenDMARC report" hour="2" minute="0" job="/bin/bash /etc/opendmarc/report.sh >> /var/log/opendmarc_report.log"
-
+  cron: name="OpenDMARC report" hour="2" minute="0" job="/bin/bash /etc/opendmarc/report.sh >> /var/log/opendmarc_report.log 2>&1 || tail /var/log/opendmarc_report.log"
--- a/roles/mailserver/templates/etc_opendmarc_report.sh.j2
+++ b/roles/mailserver/templates/etc_opendmarc_report.sh.j2
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# ensure this script errors out if any of its steps do
+set -e
+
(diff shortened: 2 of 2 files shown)


neuhaus:
Is there a way to send an email only if there is an error during the cron job? That would be ideal.

carljm:
Hmm. I think that should be possible. I'd edit the `/etc/opendmarc/report.sh` script which is run by the cron job, to make each individual step (`opendmarc-import`, `opendmarc-reports`, and `opendmarc-expire`) redirect all its "normal" output to stdout, so in a success case all the output goes to stdout. Then I'd make sure that if any step fails (nonzero exit code) it writes something to stderr telling the admin to go check the log file for details.

If that sounds like a reasonable plan, I can put together a PR for it.

carljm:
Ok, this should achieve the goal. The report script will now error out if any of its steps exit with error, and the cron job now directs all output to the log file, but in case of an error exit will dump the last few lines of the logfile in an email to root. Good enough?

carljm:
Tested and working on my server. Also, the subject of the email from cron contains the name of the log file, so it's pretty clear where to go looking for more info if there is an error.

neuhaus:
Thanks for the nice solution!

PR #517, opened 2016-03-08
Pass {auth_type} to milters, fixing OpenDKIM signing of authenticated SMTP messages.

### LINKED ISSUE #510: OpenDKIM not signing outgoing mails; doesn't think they are authenticated
When I run `vagrant up` on `jessie` branch followed by `nosetests`, the test for DKIM signature fails. Looking at syslog, I see this:

```
Mar  7 06:03:08 sovereign opendkim[17547]: 2A6AD30068B: [172.16.100.1] [172.16.100.1] not internal
Mar  7 06:03:08 sovereign opendkim[17547]: 2A6AD30068B: not authenticated
Mar  7 06:03:08 sovereign opendkim[17547]: 2A6AD30068B: no signature data
```

In `/etc/opendkim/TrustedHosts`, I see

```
127.0.0.1
10.0.2.15
sovereign.local
mail.sovereign.local
```

But from the perspective of the Sovereign VM, `sovereign.local` resolves to `127.0.0.1` (because it is listed as such in `/etc/hosts`, and `mail.sovereign.local` does not resolve at all (it is listed in the `/etc/hosts` of the host machine, but not of the Sovereign VM. So the host machine is not a trusted host for OpenDKIM, and since the tests run from the host machine, the test email is not DKIM-signed, and the test fails.

Can others reproduce this, or am I doing something wrong?

I think the simplest fix is to add the IP address of the host machine to `/etc/opendkim/TrustedHosts` when running under Vagrant. I've verified that that does in fact cause the test to pass. I can submit a PR for this if other contributors are also seeing this test failure and agree that that's the right fix.

carljm:
Fixes #510 (and fixes a failing test).

### DIFF
--- a/roles/mailserver/templates/etc_postfix_main.cf.j2
+++ b/roles/mailserver/templates/etc_postfix_main.cf.j2
@@ -104,7 +104,7 @@ local_recipient_maps = $virtual_mailbox_maps
 smtpd_milters = inet:127.0.0.1:8891,inet:127.0.0.1:54321,unix:/var/run/rmilter/rmilter.sock
 non_smtpd_milters = $smtpd_milters
 milter_protocol = 6
-milter_mail_macros = i {mail_addr} {client_addr} {client_name} {auth_authen}
+milter_mail_macros = i {mail_addr} {client_addr} {client_name} {auth_authen} {auth_type}
 milter_default_action = accept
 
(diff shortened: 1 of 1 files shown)


neuhaus:
Thanks!

PR #519, opened 2016-03-09
Remove unused variables in jessie

ariddell:
google_auth_version is not used in jessie, neither is znc_version


mikeashley:
Looks good.  Looking back at the diffs, we have switched both ZNC and google authenticator to use distribution packages instead of downloads.

Issue #394, opened 2015-06-30
Replace dspam with Rspamd


plotters:
dspam is removed as of Debian 8: https://tracker.debian.org/pkg/dspam
Is there a good replacement?

ariddell:
That's a shame. Apparently it's not maintained upstream: https://tracker.debian.org/news/558416

al3x:
That's going to be a minor PITA to swap out. Guess it's SpamAssassin for us, or maybe [Anti-Spam SMPT Proxy](https://en.wikipedia.org/wiki/Anti-Spam_SMTP_Proxy)?

neuhaus:
I found no packages for ASSP on Debian, let's pick something that has packages available and in the default distribution. Spamassassin is probably the most straightforward choice.

florianbeer:
This is really bad! Dspam was working so well for me, now I'm much more on the fence about upgrading to Jessie :(

plotters:
Any thoughts on this one:  https://www.rspamd.com/ ?

al3x:
Looks like a really strong choice. Actively maintained, good feature list, Google Summer of Code support.

al3x:
Just pushed [the beginnings of an Rspamd](https://github.com/sovereign/sovereign/commit/ecaa4c23308adbdb5d76937296090ca7a6cc5944) implementation in our `jessie` branch.

Still need the task to configure [Rmilter](https://rspamd.com/rmilter/) to be aware of [Rspamd](https://www.rspamd.com/doc/integration.html). Postfix configuration will pick up Rmilter and should "just work" from there.

carljm:
I'm working on finishing this out, and ran into a question. Apparently rmilter also has DKIM signing and validation support (using OpenDKIM under the hood). Should I disable that in the rmilter config, and use rmilter only for rspamd, and continue using OpenDKIM's own milter? Or should we just let rmilter handle both, and reduce from three down to two milters? Any opinions?

mikeashley:
We can close this since rspamd is integrated.  I created issue #529 to consider removing OpenDKIM's milter.

PR #521, opened 2016-03-09
Complete rmilter/rspamd setup.

carljm:
This gets rspamd hooked up to postfix via rmilter.

Rmilter has a bunch of other capabilities that could be explored in the future (including the ability to do DKIM signing and verification, so we wouldn't need to have Postfix talk directly to OpenDKIM as a separate milter), but this PR only sets it up for rspamd.

The `rmilter.socket` systemd unit file has to be replaced because the default unit file uses a Unix socket instead of a TCP socket, and (as discussed at https://github.com/vstakhov/rmilter/issues/39) this doesn't play nicely with postfix running smtpd in a chroot. The author of rmilter recommends just using a TCP socket instead.

### DIFF
--- a/roles/mailserver/files/lib_systemd_system_rmilter.socket
+++ b/roles/mailserver/files/lib_systemd_system_rmilter.socket
@@ -1,0 +1,8 @@
+[Unit]
+Description=Another sendmail milter for different mail checks
+
+[Socket]
+ListenStream=9900
+
+[Install]
+WantedBy=sockets.target
--- a/roles/mailserver/files/etc_rmilter.conf.common
+++ b/roles/mailserver/files/etc_rmilter.conf.common
@@ -1,0 +1,12 @@
+spamd {
+	servers = r:localhost:11333;
+	whitelist = 127.0.0.1/32, 192.168.0.0/16, [::1]/128;
+};
+
+redis {
+	servers_id = localhost;
+	id_prefix = "message_id.";
+};
+
+tempdir = /tmp;
+max_size = 10M;
--- a/roles/mailserver/tasks/rspamd.yml
+++ b/roles/mailserver/tasks/rspamd.yml
@@ -11,10 +11,26 @@
   tags:
     - dependencies
 
-- name: Install Rspamd and Rmilter
+- name: Install Rspamd, Rmilter, and Redis
   apt: pkg={{ item }} state=installed update_cache=yes
   with_items:
     - rspamd
     - rmilter
+    - redis-server
   tags:
     - dependencies
+
+- name: Configure rmilter
+  copy: src=etc_rmilter.conf.common dest=/etc/rmilter.conf.common
+
+- name: Configure rmilter socket service
+  copy: src=lib_systemd_system_rmilter.socket dest=/lib/systemd/system/rmilter.socket
+
+- name: Start redis
+  service: name=redis-server state=started
+
+- name: Start rspamd systemd socket listener
+  service: name=rspamd.socket state=started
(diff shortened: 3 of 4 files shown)


carljm:
This fixes #394.

carljm:
Oh, and this is running and tested on my new mail server; mail log shows details of successful rspamd checks of incoming messages.

roles/mailserver/files/lib_systemd_system_rmilter.socket
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Another sendmail milter for different mail checks
+
+[Socket]
+ListenStream=9900
```
ariddell: Why the switch away from unix:/var/run/rmilter/rmilter.sock ?

ariddell:
LGTM

hpk42:
hey carl, you here :)  I am still running on master and wondering if to go for the jessie branch -- may i ask if rspamd will also automatically train on inbox/spam-box so that i can move messages via an IMAP client and it gets recognized?

mikeashley:
Do we really have to intsall redis?

neuhaus:
I'd rather not add redis for this myself, however looking at its memory requirements (1MB without any data and 100MB with 1 million items) I think it's manageable.
rmilter considers memcached to be legacy software and encourages to switch to redis.

carljm:
Yeah, rmilter/rspamd seem to have a strong preference for having redis available and complains at you in its log if you don't. I don't think it's strictly required, but it's necessary for some features, including bayesian learning, which I think we want available. Personally, I don't mind having Redis around; it's very useful, it's stable, and it has a small memory footprint.

Speaking of the Bayes filter, hello @hpk42 ! :-) So you can train rspamd with `rspamc learn_ham email_file1 email_file2 ...` or `rspamc learn_spam email_file1 email_file2`. Rspamd doesn't have anything set up to do that automatically, and currently this PR doesn't either. I wasn't sure how opinionated we wanted Sovereign to be on that.

One option would be a cron job that runs learn_spam/learn_ham on some specially-named ham/spam folders (probably configurable by vars?) in the main user's maildir (I think I'd need to look into how to only run it on new messages that weren't already processed). The downside of this approach is that it doesn't scale quite as well to multi-email-user Sovereign installations (though I suppose we could have it run for all configured mail accounts).

Another option (recommended by the rspamd author) is to set up local mail aliases that pipe the mail directly to learn_spam/learn_ham. The downside of this is that you need some way to conveniently redirect mail to these aliases from your MUA; e.g. Thunderbird has a "redirect mail" plugin that will do this.

Open to thoughts on this, though I think the PR could be merged as-is to at least make rspamd functional, and any desired setup for Bayes learning could be treated as a separate enhancement.

mikeashley:
I like the idea of getting rspamd properly set up and then tweaking further in separate PRs if folks feel improvement is needed.  That said, mail setup isn't trivial.  Is there somebody who owns the mail server configuration and is watching that it's coherent?

carljm:
I haven't examined the mail server config in great detail yet, so I can't speak to its overall coherency, but I can verify that it works :) including DKIM and (with this PR) rspamd. 

In any case, I don't think there's any concern that this PR makes the mail server config any less "coherent." It just completes the rspamd work that was earlier started and left half-done. The impact on the mail server config is pretty minimal, as you can see in the diff.

The main potential lack of coherence I see is that we have two milters (opendkim and rmilter) whose tasks could both be handled by rmilter. I plan to look into that in a future PR.

PR #525, opened 2016-03-09
openvpn: only CN needed for openssl certificate

ariddell:
There is no requirement, AFAIK, that a self-signed x509 certificate contain any
information other than common name (CN). Requiring users to provide county, state, city, etc. is inconvenient and, if I have this right, unnecessary.

### DIFF
--- a/group_vars/sovereign
+++ b/group_vars/sovereign
@@ -51,11 +51,6 @@ prosody_accounts:
     password: TODO
 
 # openvpn
-openvpn_key_country:  "US"
-openvpn_key_province: "California"
-openvpn_key_city: "Beverly Hills"
-openvpn_key_org: "ACME CORPORATION"
-openvpn_key_ou: "Anvil Department"
 openvpn_clients:
   - laptop
   - phone
--- a/roles/vpn/tasks/openvpn.yml
+++ b/roles/vpn/tasks/openvpn.yml
@@ -38,7 +38,7 @@
         mode=0600
 
 - name: Generate CA certificate
-  command: openssl req -nodes -batch -new -x509 -key {{ openvpn_ca }}.key -out {{ openvpn_ca }}.crt -days {{ openvpn_days_valid }} -subj "{{ openssl_request_subject }}/CN=ca-certificate"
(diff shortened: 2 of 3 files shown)


mikeashley:
I like the change.  Have you tested it?  Also, I don't know a lot about certificates, so it would be good if somebody else could comment on the wisdom of this change.

ariddell:
I do need to test this with a laptop or some other device.

vars/defaults.yml
```diff
@@ -84,13 +84,10 @@ tarsnap_version: 1.0.36.1
...
 # Notes about security: https://blog.g3rt.nl/openvpn-security-tips.html
 # Check privacy: http://witch.valdikss.org.ru/
-# openvpn_key_country: (required)
-# openvpn_key_province: (required)
-# openvpn_key_city: (required)
-# openvpn_key_org: (required)
-# openvpn_key_ou: (required)
 openvpn_days_valid: "1825"
-openssl_request_subject: "/C={{ openvpn_key_country }}/ST={{ openvpn_key_province }}/L={{ openvpn_key_city }}/O={{ openvpn_key_org }}/OU={{ openvpn_key_ou }}"
+# if you would  like to define your country, province, and locality, use the following
```
ariddell: Assuming things work I don't even think this comment is needed.

mrwacky42:
This would potentially break #561 with respect to verifying the server's x509 cert.
I'd rather just have dummy defaults and suggest (however weakly) that the user supply them.

ariddell:
If/when #526 gets merged there's some cleanup that needs to happen in this PR. These lines need to be removed from group_vars/sovereign
```
#openvpn_key_country:  "US"
#openvpn_key_province: "California"
#openvpn_key_city: "Beverly Hills"
#openvpn_key_org: "ACME CORPORATION"
#openvpn_key_ou: "Anvil Department"
```

ariddell:
Verified working on a Debian laptop.

mikeashley:
@ariddell #526 was merged.  Can you confirm whether this PR is ready to go or update it?

ariddell:
Sure. I'll rebase it.

PR #526, opened 2016-03-09
Modularize sovereign

ariddell:
**This is a major change to sovereign**

This modularizes sovereign by moving default variables into their roles. It responds to demands for a less monolithic sovereign setup expressed in #440 and elsewhere.

Highlights:

- Users need only define two variables for all roles to work: ``domain`` and ``main_user_name``
- All passwords are generated using ansible's [password lookup](https://docs.ansible.com/ansible/playbooks_lookups.html#the-password-lookup) and stored in the directory `secret`.
- Uses the ansible standard ``group_vars/sovereign`` instead of the bespoke ``vars/user.yml``
- Default variables are moved into their respective roles (``defaults/main.yml``)

Documentation will be added and testing will be fixed if there's interest in this change.


README.md
```diff
@@ -72,6 +72,9 @@ Create a new machine key for your server:
 
 ### 2. Prep the server
 
+Some servers allow the root user to login. If your server is such a server, you
```
carljm: I'm not sure this sentence adds any value. It's not necessary, because changing the root password is a reasonable recommendation regardless of whether you allow root logins. And it's confusing in this location, because it seems to apply to all the steps under "prep the server," but it doesn't -- you need to create an account for Ansible regardless, and authorizing your ssh key is unrelated to whether you allow root logins.

ariddell: Sorry, you're right, I need to move this out of the PR

carljm:
This looks great! I think it's a big improvement, thanks for putting it together.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .vagrant
 vagrant_ansible_inventory_default
 tests.pyc
+secret
```
carljm: I wouldn't personally want this directory git-ignored. I consider an Ansible deploy repo (as opposed to a reusable role) to be private, and I treat it that way. I want my important secrets tracked in git, not subject to deletion by a stray `git clean`.

But if others prefer to have it git-ignored, I can just revert that change in my private fork for my server.

ariddell: I think the idea here is that you're free to ``git add -f`` so your
secrets are tracked but the sensible default is to prevent
them from being added by accident.

ariddell:
Anyone else have thoughts? If there's interest in this I'll go ahead and add more documentation.

roles/vpn/templates/openssl-server-certificate.cnf.j2
```diff
@@ -39,19 +39,14 @@ distinguished_name = req_distinguished_name
 
 [ req_distinguished_name ]
 countryName = Country Name (2 letter code)
-countryName_default = {{ openvpn_key_country }}
```
Yannik: What happened here?

Yannik: Oh, got it!

Yannik:
@ariddell I have looked through all your changes. This would be a great improvement to sovereign.

mikeashley:
I want to spend some time looking at this.  I still have my doubts but 
am being open-minded about it.  That said, I am swamped with work, and 
it may be another week before I can look at it.

roles/vpn/templates/openssl-server-certificate.cnf.j2
ariddell: My mistake. I comingled #525 in this PR. I'll fix this.

Yannik:
Thanks for removing the openvpn part, great work, @ariddell!

ariddell:
I think it would be great to merge this -- if it gets merged at all -- before the move to jessie. The benefits of modularizing the roles (mainly, moving role-specific variables into roles) are easy to understand. The downside is the lack of backwards compatibility.

roles/common/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+common_timezone: 'Etc/UTC'
+
+# encfs
+encfs_password: "{{ lookup('password', 'credentials/' + role + '/encfs_password' length=32') }}"
```
mikeashley: How do you envision migrating existing passwords?  Also, did you consider using Ansible Vault?  The ansible docs for `lookup` suggested that Vault may be the better option.

roles/news/handlers/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: restart apache
```
mikeashley: For hygiene suggest splitting this patch into a separate commit.

group_vars/sovereign
```diff
@@ -0,0 +1,66 @@
+# required variables
+# ------------------
+# only two variables *must* be set:
```
mikeashley: Isn't this a little misleading?  I think your intent is that a user reviews everything in this file and updates as needed.  Email passwords should be set, for example.

roles/ircbouncer/tasks/znc.yml
```diff
@@ -42,9 +42,13 @@
...
   file: state=directory path=/usr/lib/znc/configs group=znc owner=znc
 
+- name: Notify user that irc_nick must be defined
+  debug: msg="`irc_nick` and related variables must be defined for znc to work."
+  when: irc_nick is not defined
+
 - name: Copy znc configuration file into place
   template: src=usr_lib_znc_configs_znc.conf.j2 dest=/usr/lib/znc/configs/znc.conf owner=znc group=znc
-  when: znc_config.rc != 0
+  when: znc_config.rc != 0 and irc_nick is defined
```
mikeashley: It seems odd that you are trying to cover a misconfiguration error here when we don't in other places, e.g., prosody setup and mailserver setup.

mikeashley:
I think this is a good step.  My biggest question is around the password lookup method and how we would migrate existing passwords.  It moves role-specific variables into their roles and makes it a little easier to reason about changes to them.

Unless I'm mistaken, this is only a small part of addressing resuability?  As written today, some roles still make assumptions about how other roles are designed and used.  Two examples: our LE setup and duplicated postgres install/configuration.  I don't think this PR needs to full address resuability, but I want to make sure we're clear that this is only one step.

roles/common/defaults/main.yml
ariddell: Good question. The first assumption is that this only goes in jessie and
so there's some likelihood people will generate fresh passwords for
everything.

As for migration I think you can run the script once and then edit the
files in credentials/ that Ansible creates.

I did consider Vault. There's also a host of other alternatives too,
like Mozilla's [sops](https://github.com/mozilla/sops),
[credstash](https://docs.ansible.com/ansible/playbooks_lookups.html#the-credstash-lookup),
etc.

I avoiding shifting to Vault because I think that probably is a separate
discussion. No secrets are encrypted now and this commit doesn't change
it. (I'd be open to switching to Vault.)

ariddell: Why is this marked as an "outdated diff"?

ariddell:
(Line comments have been addressed)

This is indeed "only a small part of addressing resuability", as @mikeashley puts it. The title of the PR is perhaps too ambitious. I would love to do more.

I find that inter-dependencies like postgres are difficult to deal with in ansible. I haven't seen a satisfactory solution really (although some of @drybjed's tricks in https://github.com/debops/debops are interesting). My sense is you need some sort of key-value store to register what roles have been successfully run. I'm not sure sovereign's quite ready for that. It might be easier, in practice, to remove the inter-dependencies with role-specific databases.

drybjed:
@ariddell You are exactly right, some way to store a state in Ansible (which is inherently stateless by design) so that it can be carried over between the playbook executions, either in the same `ansible-playbook` run, or on subsequent runs, is needed to get full idempotency. In DebOps, I'm doing this using Ansible local facts stored on remote hosts (a JSON data structure with key/value data relevant to a particular role).

In the case of PostgreSQl support, which is done by the [debops.postgresql_server](http://docs.debops.org/en/latest/ansible/roles/ansible-postgresql_server/docs/getting-started.html) role on the server side, and [debops.postgresql](http://docs.debops.org/en/latest/ansible/roles/ansible-postgresql/docs/getting-started.html) role on the client side, I use Ansible local facts to pass the relevant data between two roles. The default variables are not good for this, because when a role is not included in the play, its variables are not "loaded" by Ansible. Other forms of variables available in Ansible are either not modifable through Ansible inventory, or the inventory itself which should be left under user control.

If you wonder why there are two PostgreSQL roles (and MariaDB, and similar), this allows an easy setup where the database server is on one host and clients that access the database are on another host(s).

As for remembering what roles have been executed before... I don't think that's really needed. You just need to ensure that each role is idempotent, and using the playbook, run your roles in the correct order. For example, my application roles don't really care if `nginx` role has been executed or not, they expect the webserver to just be there when it's needed.

mikeashley:
I recommend we take this PR, but it would be good to get a look from one more committer.  @neuhaus or @jplock or @lukecyca?

Documentation still needs updated.  Also, has this been tested?  @ariddell are you sure about password migration?  People upgrading their Debian 7 servers would need to do this.

ariddell:
I have tested this. I'll do some additional testing about the password
migration when I write the docs.

neuhaus:
I'm in favor of this PR, however I don't have time to test it right now.

ariddell:
Thanks. In that case, I'll work on the documentation.

mikeashley:
@ariddell Are you ready for this to be merged? If so, I'll do a final review and likely do some testing.

ariddell:
It's ready except for documentation updates. I'm going to try to test it
fully this week and write the docs.

mrwacky42:
This looks like more idiomatic Ansible code. The way vars are set up currently took me a few minutes to figure out.

.gitignore
QWp6t: Shouldn't you be using [ansible-vault](http://docs.ansible.com/ansible/playbooks_vault.html) for secrets? Then you can safely commit them to the repo.

ariddell:
This is ready to be merged.

mikeashley:
I'll test it, taking it from the top.  I'll consider @QWp6t's input on ansible-vault as well as I work my way through a test setup.

mikeashley:
Regarding secrets, I like the way @ariddell has done it.  It makes the secrets available to me in text files on my local machine without any fanfare.  I can commit them to my private branch.  I can easily copy them to use when logged into my server.

I tested all roles except tarsnap.  The variable `irc_timezone` got dropped accidentally.  I submitted a PR to @ariddell to correct it.  Once this PR is updated, I'll merge it.  We'll need some guidance on the wiki on how to migrate passwords in the .yml files to text files in the `secret` directory.

ariddell:
Merged the PR. I can't take credit for the secrets approach, it's a scaled-down version of @drybjed's approach in debops.

PR #528, opened 2016-03-11
Only add iptables rules if they don't already exist

Yannik:


mikeashley:
Thank you for the care you are putting into this!

PR #531, opened 2016-03-12
Use https for rpsamd key and repository

spk:
Use secure https
Cheers


carljm:
+1. This was my oversight when updating the rspamd source handling. (I think I was just following the rspamd docs, which use http.) Thanks for the cleanup!

Issue #509, opened 2016-03-07
Don't bounce Apache on certificate renewal


mikeashley:
Currently, LE certificate generation bounces Apache.  This isn't required if webroot authentication is used.  The common role should be updated to use this method.

PR #533, opened 2016-03-15
Idempotency fixes to LetsEncrypt handling.

carljm:
Previous to this PR, because the ZNC and dovecot roles were modifying the LetsEncrypt renewal cron job in place, every subsequent run would reinstall the original version of that cron job. This causes spurious "changed" tasks, as well as problems if e.g. the LetsEncrypt tasks are run outside of a full run (because the cronjob will be restored to its non-updated form). This PR fixes that by making each post-renew task an independent script, found and executed by the cronjob. Tested and working.

### DIFF
--- a/roles/common/files/etc_cron-monthly_letsencrypt-renew
+++ b/roles/common/files/etc_cron-monthly_letsencrypt-renew
@@ -18,5 +18,8 @@ for c in $(find /etc/letsencrypt/live/ -mindepth 1  -type d); do
 done
 service apache2 start
 
-# Services that rely on LE certificates will need restarted.
-
+# Services that rely on LE certificates may need restarted and/or other actions.
+for script in $(find /etc/letsencrypt/postrenew/ -maxdepth 1 -type f); do
+  echo "Executing ${script}."
+  $script
+done
--- a/roles/common/files/etc_cron-monthly_letsencrypt-renew
+++ b/roles/common/files/etc_cron-monthly_letsencrypt-renew
@@ -19,7 +19,7 @@ done
 service apache2 start
 
 # Services that rely on LE certificates may need restarted and/or other actions.
-for script in $(find /etc/letsencrypt/postrenew/ -maxdepth 1 -type f); do
+for script in $(find /etc/letsencrypt/postrenew/ -maxdepth 1 -type f -executable); do
   echo "Executing ${script}."
   $script
 done
--- a/roles/common/files/letsencrypt-gencert
+++ b/roles/common/files/letsencrypt-gencert
@@ -1,4 +1,5 @@
 #!/bin/bash
+service apache2 stop
 d="$1"
 for i in www mail autoconfig read news cloud git; do
   if (getent hosts $i.$1 > /dev/null); then
--- a/roles/common/files/letsencrypt-gencert
+++ b/roles/common/files/letsencrypt-gencert
@@ -5,6 +5,9 @@ for i in www mail autoconfig read news cloud git; do
     d="$d,$i.$1";
   fi
 done
+# We are using the "standalone" letsencrypt plugin, which runs its own
+# webserver, so we need to temporarily free up the HTTP(S) ports by stopping
+# our own Apache.
 service apache2 stop
 /root/letsencrypt/letsencrypt-auto certonly -c /etc/letsencrypt/cli.conf --domains $d
 service apache2 start
--- a/roles/common/files/letsencrypt-gencert
+++ b/roles/common/files/letsencrypt-gencert
@@ -1,9 +1,10 @@
 #!/bin/bash
-service apache2 stop
 d="$1"
 for i in www mail autoconfig read news cloud git; do
   if (getent hosts $i.$1 > /dev/null); then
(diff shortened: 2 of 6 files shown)


carljm:
Added another commit to this, which just makes the LetsEncrypt dependencies installation report its "changed" status correctly.

carljm:
...and one more commit, fixing the cert-request task to actually not request a new cert if we already have one. Seems the `args:` form of passing args may only work with the "command" module, not the "script" module? Not sure, but it wasn't working, and this form does work. (Ansible 1.9.3).

carljm:
...and one last fix, to avoid bouncing Apache unless we are actually going to generate a new cert.

At this point, the `letsencrypt.yml` tasks run clean and unchanged, and don't do anything unnecessary, when nothing needs to be done.

roles/common/files/etc_cron-monthly_letsencrypt-renew
```diff
@@ -18,5 +18,8 @@ for c in $(find /etc/letsencrypt/live/ -mindepth 1  -type d); do
 done
 service apache2 start
 
-# Services that rely on LE certificates will need restarted.
-
+# Services that rely on LE certificates may need restarted and/or other actions.
+for script in $(find /etc/letsencrypt/postrenew/ -maxdepth 1 -type f); do
```
neuhaus: I suggest you add "-executable" to the find options, that way we can add files such as a README without the execute bit and they won't be executed

roles/common/files/letsencrypt-gencert
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+service apache2 stop
```
neuhaus: This can be moved after the for loop.
Where is apache2 started again?

roles/common/files/etc_cron-monthly_letsencrypt-renew
carljm: Good idea; done.

roles/common/files/letsencrypt-gencert
carljm: Moved to after for loop, good point.

I had debated about whether to start Apache again in this script. It is started by a later task in the same `letsencrypt.yml` tasks file. I hadn't re-started it here because I wasn't sure that in all cases (e.g on first run) it would be started yet here in the first place. But actually I think it will be (because it's an enabled service by default), and in any case there's no harm in starting it, and it's clearer to have a paired start/stop within this script. So I added a start at the end.

mikeashley: Should the start of Apache now be deleted from the end of the `letsencrypt.yml` task file?

carljm: No, I don't think so. This gencert script runs only conditionally, if there isn't an existing cert. Regardless, we want to be sure that Apache is running, and we can let Ansible take care of that. If it's already running, the Ansible task won't do anything, so there's no harm done.

mikeashley: I am thinking about the invariant.  Apache is running when this task list runs.  After the gencert script (which is run conditionally), Apache is left running.  I don't see the need for the last start task, and I worry it will confuse future readers who also wonder why it's there, trying to answer the question, "what is the condition that led to the author thinking they needed to (re)start Apache from Ansible???"

neuhaus:
I'm confused about `roles/common/files/letsencrypt-gencert`: The script stops apache, then runs the letsencrypt client with `certonly`, then starts apache again.

Doesn't the `certonly` option require apache to be running?

carljm:
@neuhaus This PR doesn't change the fact that Apache is stopped before running the letsencrypt client, it just moves that logic directly into the `gencert` script instead of having it in Ansible.

The LE config that @mikeashley set up uses the `standalone` letsencrypt plugin, which runs its own webserver, thus it needs ports 80/443 available, thus we have to shut down our Apache first.

It might be worth considering whether we should instead use the `apache` or `webroot` plugin, and have our own Apache instance serve the LetsEncrypt confirmation stuff. Then we wouldn't need to bounce it. I'm guessing @mikeashley might have more insight as to why he didn't choose that route.

See http://letsencrypt.readthedocs.org/en/latest/using.html#plugins for more on the available LE plugins.

I can (and will) add a comment into the `gencert` script clarifying the above.

roles/common/files/letsencrypt-gencert
carljm: @mikeashley Yeah, on second thought you're right. I was thinking that Apache was somehow intrinsically related to this set of tasks, but really it isn't; there's no particular reason for this task file to be responsible for ensuring that it's running. I'll remove it.

carljm:
Ok, I think I've addressed all the comments here; take another look?

mikeashley:
I think these are good changes and should be merged.

Issue #509 suggests not bouncing Apache.  We agreed this is the right thing to do.  We're just waiting for a patch from me or somebody else to make it happen.

mikeashley:
I checked this on my test server.  So nice to see `changed=0`.  Thank you!

carljm:
:+1:

PR #534, opened 2016-03-15
Workaround bug with enabling SysV init scripts on Jessie.

carljm:
Without this PR in Jessie, the `enabled=yes` arg on the ntp service task has no effect and always reports `changed: true`, due to https://github.com/ansible/ansible-modules-core/issues/915

This PR works around that bug to actually ensure that NTP is enabled and fix the changed reporting.

### DIFF
--- a/roles/common/tasks/ntp.yml
+++ b/roles/common/tasks/ntp.yml
@@ -11,6 +11,10 @@
   notify:
     - restart ntp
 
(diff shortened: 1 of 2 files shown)


carljm:
Also worked around the same bug for `collectd`.

neuhaus:
Thanks!

PR #535, opened 2016-03-15
Encrypt Postgres passwords, and fix change-reporting.

carljm:
This causes the Postgres role passwords to be encrypted rather than plaintext in the database.

It also uses the `postgresql_user` Ansible module to apply the change to the admin user's password; this fixes changed-reporting. (The previous version of that task always reported `changed: true`).

### DIFF
--- a/roles/mailserver/tasks/postfix.yml
+++ b/roles/mailserver/tasks/postfix.yml
@@ -16,7 +16,7 @@
   tags:
     - dependencies
 
-- name: Set postgres user password
+- name: Set password for PostgreSQL admin user
   become: true
   become_user: postgres
   postgresql_user: name={{ db_admin_username }} password={{ db_admin_password }} encrypted=yes
--- a/roles/mailserver/tasks/postfix.yml
+++ b/roles/mailserver/tasks/postfix.yml
@@ -16,12 +16,14 @@
   tags:
     - dependencies
 
-- name: Set postgres password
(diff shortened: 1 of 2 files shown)


roles/mailserver/tasks/postfix.yml
```diff
@@ -16,12 +16,14 @@
   tags:
     - dependencies
 
-- name: Set postgres password
-  command: sudo -u {{ db_admin_username }} psql -d {{ db_admin_username }} -c "ALTER USER postgres with  password '{{ db_admin_password }}';"
+- name: Set postgres user password
```
neuhaus: the comment is misleading, it sets the admin password. How about:
"Set password for ProgreSQL admin user"

carljm: Done.

mikeashley:
ownCloud configures the postgres admin user also.  Could this PR address that also with a similar change?  Should we be assuming that all users will use postfix?

neuhaus:
Yeah, I don't see anyone switching from Postfix to something else at the moment.

carljm:
Ok, I made the same change to the owncloud role's configuration of the PG admin user.

Seems like really that task ought to be centralized into a PG role rather than repeated in each role that uses PG, but that's beyond the scope I want to get into in this PR. It gets into the modularity questions that are being discussed elsewhere.

I think this is ready for merge.

mikeashley:
Looks good to me.  @neuhaus, any other comments?

neuhaus:
Looks good to me as well. Thanks for the patch.

PR #536, opened 2016-03-15
Idempotency/changed-reporting fixes for OpenDMARC tasks.

carljm:
As discussed in https://groups.google.com/forum/#!msg/ansible-project/XrQ08TdtcRk/I4s7rF7GkyMJ the `mysql_db` module's `import` arg unconditionally re-runs the given SQL file, whether needed or not, and always reports `changed: true`. This PR only imports the schema after initially creating the database.

Similarly, the `file` module's `touch` arg updates the mtime and atime (which we don't need here) and always reports `changed: true`. Using `copy` instead with empty contents ensures the file exists with given ownership, but only returns `changed: true` if it was actually created or its ownership was changed.

With these changes, the OpenDMARC tasks run changed-clean for me.

### DIFF
--- a/roles/mailserver/handlers/main.yml
+++ b/roles/mailserver/handlers/main.yml
@@ -19,3 +19,6 @@
 
 - name: restart rspamd
   service: name=rspamd.socket state=restarted
+
+- name: import opendmarc schema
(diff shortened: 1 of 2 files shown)


mikeashley:
Looks good.  Thanks!

PR #541, opened 2016-03-27
Fix monitoring config for apache/postgresql.

johngian:
I did a clean installation of a new Debian jessie server configured using sovereign. It looks like monit configuration for pgsql/apache doesn't check the correct process ``pid`` files.


roles/monitoring/files/etc_monit_conf.d_pgsql
```diff
@@ -1,4 +1,4 @@
-check process postgres with pidfile /var/run/postgresql/9.1-main.pid
+check process postgres with pidfile /var/run/postgresql/9.4-main.pid
```
neuhaus: On Ubuntu Xenial, both PostgreSQL 9.4 and 9.5 are available. I suppose we stick to 9.4 to improve compatability with Debian Jessie

neuhaus:
Thanks!

PR #542, opened 2016-03-30
Add directories to allow persistent znc user/modules configuration.

johngian:


johngian:
The configuration generated by sovereign regarding znc doesn't create some folders required by znc. This causes all the user/modules changes and content not to be persistent in the disk.

neuhaus:
I don't use ZNC myself. 
What do you think about using per-user files for ZNC in `~/.znc/`?

johngian:
We are using a custom ``datadir`` when running the znc service:
``/usr/bin/znc --datadir=/usr/lib/znc``

That means that the path where znc saves everything is going to be ``datadir``.

Also we are creating a user/group named ``znc`` to spawn the ``znc`` daemon and this user's home is ``/usr/lib/znc``. That means that ``~/.znc`` will be ``/usr/lib/znc/.znc`` which doesn't really make sense because of ``datadir``.

neuhaus:
OK, thanks for clarifying and thanks for the pull request.

PR #544, opened 2016-04-09
Add non-interactive to LE client options

mikeashley:
Depending on when the client is run, there are no certificates to
update.  By default, the client runs in interactive mode and wants to
notify the user of this.  This causes Ansible to hang waiting for an
acknowledgement that will never come.  Adding the non-interactive flag
fixes this.

### DIFF
--- a/roles/common/templates/etc_letsencrypt_cli.conf.j2
+++ b/roles/common/templates/etc_letsencrypt_cli.conf.j2
@@ -5,3 +5,4 @@ register-unsafely-without-email = True
 keep = True
 expand = True
 agree-tos = True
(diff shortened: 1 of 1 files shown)


mikeashley:
I cannot reproduce the CI build errors from `ansible-lint` but am running a valid version.  Is anybody else seeing this?

carljm:
I do get those errors locally, not only on this branch but on the main `jessie` branch too. It probably depends on ansible-lint version? I have 2.6.1.

If it is just a new warning in a more recent ansible-lint, it should be fixed, but also doesn't need to hold back this PR.

PR #545, opened 2016-04-09
Use the Apache interface to disable a site

mikeashley:


mikeashley:
This is a bad patch.  It only works if ownCloud is already installed.

Issue #524, opened 2016-03-09
Optional support for www.{{domain}}


ariddell:
Heroku recommends against using A records and "naked domains"

> DNS A-records can be used to configure zone apex domains (also called bare, naked, or root domains). These records have serious availability implications when used in environments such as on-premise data-centers, cloud infrastructure services, and platforms like Heroku.

There's also this site: http://www.yes-www.org/

This all makes sense and if a sovereign user would like to force HTTP/S requests to go to ``www.{{domain}}`` this should be supported.

neuhaus:
:+1: 
On my server I also redirect from 
http://domain to https://www.domain and
https://domain to https://www.domain

neuhaus:
We don't want to redirect the ACME challenges, so they must not be redirected. Here's one of the VirtualHost blocks for an site that needs to be redirected to the version with HTTPS and a "www" prefix. It's not pretty but it works:

    <VirtualHost *:80>
            ServerName example.com
            Alias /.well-known/acme-challenge/ /var/www/challenges/
            <Directory /var/www/challenges>
                    Order allow,deny
                    Allow from all
            </Directory>

            RedirectMatch permanent "^(?!\/\.well-known\/acme-challenge\/)(.*)" "https://www.example.com$1"
    </VirtualHost>

PR #548, opened 2016-04-17
add www.domain DNS entry to README.md

Dosenpfand:


mikeashley:
What error do you see if this is omitted?  You shouldn't need it if you don't include the blog role.

Dosenpfand:
I get a DNS  error by letsencrypt, because the entry does not exist:
> Domain: www.mydomain.com
> Type:   connection
> Detail: DNS problem: NXDOMAIN looking up A for www.mydomain.com

Upon closer inspection I think it should not fail otherwise since [letsencrypt-gencert](https://github.com/sovereign/sovereign/blob/jessie/roles/common/files/letsencrypt-gencert) checks the hosts existense first. I don't know why it failed in my case.
Nevertheless I think the README should include the www entry, maybe with it's role.

ariddell:
#524 is related. What if we just skipped the naked domain entirely and shifted the blog to www.domain?

mikeashley:
I'm not sure it's related.  Root cause on this one still hasn't been identified.

mikeashley:
@Dosenpfand This is a good update to the README.  Please squash the commit, and I'll merge it in.

Dosenpfand:
@mikeashley done

PR #551, opened 2016-04-28
Fix fail2ban dovecot config.

carljm:
The fail2ban config for dovecot was referring to a log file `/var/log/maillog`, which does not exist. It should be `/var/log/mail.log`. This causes fail2ban to fail to start. I only noticed because that in turn caused the post-rotate script for fail2ban's logrotate config to error out.


carljm:
This PR included commits it should not have; it is replaced by #552. Sorry for the noise.

(All info in this PR which should not have been included was immediately changed.)

PR #553, opened 2016-05-01
LE options fix

mikeashley:
LE client no longer accepts lists of domains for certificate renewal and
instead renews all installed certificates due to be renewed.  This is
the behavior we want, and it simplifies the script.


mikeashley:
Rescinded.  PR #554 taken instead.

PR #555, opened 2016-05-03
Remove unneeded/deprecated apache configuration

Dosenpfand:
- NameVirtualHost is [not needed anymore](https://httpd.apache.org/docs/2.4/mod/core.html#namevirtualhost) since version 2.3.11
- SSLCertificateChainFile is [deprecated](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslcertificatechainfile), SSLCertificateFile was extended to also load intermediate CA certificates from the server certificate file since version 2.4.8. See also Let's Encrypt [user guide](https://letsencrypt.readthedocs.io/en/latest/using.html#where-certs).


ariddell:
Thanks!

Issue #374, opened 2015-05-30
Fix Postfix submission port configuration


Yannik:
I noticed that the submission protocol is disabled in [master.cf](https://github.com/al3x/sovereign/blob/master/roles/mailserver/files/etc_postfix_master.cf).

According to [RFC6409](https://tools.ietf.org/html/rfc6409), the submission protocol / port 587 is 
  - the port which should be used for mails from a MUA (mail user agent)
  - must require authentication
  - for client->server communication.

Port 25/465 are for message relay, which is server->server communication. Port 465 uses SSL/TLS before the smtp communication starts, while 25 shall promote starttls via smtp protocol which may be used to upgrade the connection to use SSL/TLS. Port 465/ssmtp has been deprecated in 1998(!).

Even though this settings seems to exist since the first commit to the project, there has been a bit of discussion about this: https://github.com/al3x/sovereign/pull/318
This discussion is ended by stating that STARTTLS is less secure then using ssmtp, because ISPs previously tampered with the STARTTLS flag. In this case, this is untrue the submission protocol/port 587 should definitely be used. By using `smtpd_tls_auth_only = yes` in [master.cf](https://github.com/al3x/sovereign/blob/master/roles/mailserver/templates/etc_postfix_main.cf.j2#L46), it is already ensured that clients can only authenticate when using tls, mitigating any tampering from ISPs.

I happened to stumble upon this when looking through your project, but don't actually use it, and therefore don't have a complete overview of your config. Because of this, even though I'm a professional mailserver admin, I ask you to create a PR for this yourself keeping your complete config in mind.

Yannik:
Your [autoconfig](https://github.com/al3x/sovereign/blob/master/roles/mailserver/templates/var_www_autoconfig_mail_config-v1.1.j2) must be adjusted accordingly.

ariddell:
@Yannik Thanks for taking the time to leave these useful reports.

al3x:
Yes indeed! Lots to work on in the coming days.

neuhaus:
@Yannik Thank you for opening this issue.
If there is an active man-in-the-middle attack (such as stripping STARTTLS), the clients must be configured to enforce STARTTLS, otherwise the attacker can talk plain text to the client and use STARTTLS while talking to the server.

That said, I think adding the submission port is The Right Thing.

Yannik:
@neuhaus Yeah, that's true. We need something like HSTS for smtp :)

ariddell:
[...]

PR #560, opened 2016-05-21
Use submission port for client outgoing email

mikeashley:
Currently client email is submitted on port 465.  This has been
deprecated for years.  The correct way to submit email is on the
submission port (587).

This patch makes this change.  The postfix master.cf file is modified to
use the submission port, carrying over relevant configuration from the
smtps port.  Both firewall rules and the email autoconfig file are
updated accordingly.

### DIFF
--- a/roles/mailserver/files/etc_postfix_master.cf
+++ b/roles/mailserver/files/etc_postfix_master.cf
@@ -13,12 +13,13 @@ smtp       inet  n       -       -       -       1       postscreen
 smtpd      pass  -       -       -       -       -       smtpd
 dnsblog    unix  -       -       -       -       0       dnsblog
 tlsproxy   unix  -       -       -       -       0       tlsproxy
-#submission inet  n       -       -       -       -       smtpd
-#  -o syslog_name=postfix/submission
-#  -o smtpd_tls_security_level=encrypt
-#  -o smtpd_etrn_restrictions=reject
-#  -o smtpd_client_restrictions=permit_sasl_authenticated,reject
-#  -o milter_macro_daemon_name=ORIGINATING
+submission inet  n       -       -       -       -       smtpd
+  -o syslog_name=postfix/submission
+  -o smtpd_tls_security_level=encrypt
+  -o smtpd_etrn_restrictions=reject
+  -o smtpd_client_restrictions=permit_sasl_authenticated,reject_unauth_destination,reject
+  -o smtpd_sasl_security_options=noanonymous,noplaintext
+  -o smtpd_sasl_tls_security_options=noanonymous
 
 # SMTP over SSL/TLS on port 465.
 smtps     inet  n       -       -       -       -       smtpd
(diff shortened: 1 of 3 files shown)


mikeashley:
This addresses #374.  Please review.  Specific questions:

* I did not carry over `wrappermode` based on [this guidance](http://www.postfix.org/TLS_README.html#server_enable).  Ok?
* My update to `autoconfig.xml` specifies STARTTLS with the expectation the server will require TLS because `security_level` is `encrypt`.  Ok?

mikeashley:
I tested the postfix/firewall change.  I have not tested the autoconfig change but checked an example online, so I believe it's right.

JasperWallace:
Although 465 is depreciated there has been a recently increase in desire to use it - the problem with 587 is that people can do mitm attacks to block STARTTLS, and some clients will stop doing TLS at that point, which is not desirable!

465 on the other hand is TLS only.

There's no harm in doing both.

mikeashley:
This patch configures 587 to be TLS only.  It does send STARTTLS but requires the client to request a TLS connection.  According to the docs, the `wrappermode` semantics are a bit different but give you TLS also.

Indeed, there is no harm in doing both, and doing both would make this change backwards-compatible with our existing setup.  Any disagreement on enabling both?

mikeashley:
I updated this PR to leave port 465 open, and the commit message is reworded.  I believe this is ready to go if somebody can take a final look.

neuhaus:
Looks good to me.
The only thing that's missing is a small patch to the [roundcube config file](https://github.com/sovereign/sovereign/blob/master/roles/webmail/files/etc_roundcube_main.inc.php#L150) so it uses the port as well.

ariddell:
@Yannik do you have any thoughts about @JasperWallace's comment and/or this PR?

aminb:
I was testing this along with the roundcube PR, and when I clicked 'Send', this error poped up:

> SMTP Error (-1): Connection to server failed.

And here's the error logs:

```
[04-Jun-2016 21:18:00 UTC] ERROR: fsockopen(): unable to connect to ssl://mail.DOMAIN.com:587 (Unknown error) (0)
[04-Jun-2016 21:18:00 UTC] ERROR: Failed to connect socket: fsockopen(): unable to connect to ssl://mail.DOMAIN.com:587 (Unknown error) ()
[04-Jun-2016 21:18:00 +0000]: <aorclvrs> SMTP Error: Connection failed: Failed to connect socket: fsockopen(): unable to connect to ssl://mail.DOMAIN.com:587 (Unknown error) in /usr/share/roundcube/program/lib/Roundcube/rcube.php on line 1712 (POST /?_task=mail&_unlock=loading1465075081985&_lang=en&_framed=1?_task=mail&_action=send)
```

I double checked and the submission port was open as expected, and I had updated the roundcube configuration to use port `587` instead of `465`.

mikeashley:
I have not updated PR #507.  In addition to changing the port number, I think you'll need to use `tls://{{ mail_server_hostname }}` instead of `ssl://...` for the smtp server in roundcube's configuration file.

aminb:
Oh very true, I forgot to change the prefix from `ssl://` to `tls://`, and changing that (along with the port of course) did it and it's working fine. Looking forward to seeing these PRs merged :)

PR #561, opened 2016-05-24
Some VPN security enhancements

mrwacky42:
A couple more tips from https://blog.g3rt.nl/openvpn-security-tips.html

* tls-min-version (on client side)
* remote-cert-eku
* verify-x509-name

### DIFF
--- a/roles/vpn/templates/etc_openvpn_server.conf.j2
+++ b/roles/vpn/templates/etc_openvpn_server.conf.j2
@@ -323,3 +323,5 @@ sndbuf 0
 rcvbuf 0
 push "sndbuf 393216"
 push "rcvbuf 393216"
+
+remote-cert-eku "TLS Web Client Authentication"
--- a/roles/vpn/templates/openssl-server-certificate.cnf.j2
+++ b/roles/vpn/templates/openssl-server-certificate.cnf.j2
@@ -64,3 +64,36 @@ subjectKeyIdentifier=hash
 authorityKeyIdentifier=keyid,issuer:always
 extendedKeyUsage=serverAuth
 keyUsage = digitalSignature, keyEncipherment
+
+[ usr_cert ]
+
+# These extensions are added when 'ca' signs a request.
+
+# This goes against PKIX guidelines but some CAs do it and some software
+# requires this to avoid interpreting an end user certificate as a CA.
+
+basicConstraints=CA:FALSE
+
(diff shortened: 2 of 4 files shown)


roles/vpn/templates/openssl-server-certificate.cnf.j2
```diff
@@ -64,3 +64,36 @@ subjectKeyIdentifier=hash
 authorityKeyIdentifier=keyid,issuer:always
 extendedKeyUsage=serverAuth
 keyUsage = digitalSignature, keyEncipherment
+
+[ usr_cert ]
```
mrwacky42: This section straight from easy-rsa examples..

roles/vpn/templates/etc_openvpn_server.conf.j2
```diff
@@ -323,3 +323,5 @@ sndbuf 0
 rcvbuf 0
 push "sndbuf 393216"
 push "rcvbuf 393216"
+
+remote-cert-eku "TLS Web Client Authentication"
```
ariddell: Is this misleading? VPN != Web

ariddell:
My sense is that none of this is needed. We are distributing certificates. There's no need to check CN or extended key usage.

roles/vpn/templates/etc_openvpn_server.conf.j2
mrwacky42: It certainly is, but that's how TLS Extended Key Usage names it.

mrwacky42:
The justification for tips 1 & 2 is here: https://blog.g3rt.nl/openvpn-security-tips.html.
Just more extra armor to make sure you aren't being MITM'd

ariddell:
I don't think there's a MITM threat here. The client has a copy of the
server certificate on disk and that server certificate is self-signed.
Am I missing something?

ariddell:
Our server certificate is self-signed and we distribute the CA certificate so there's no MITM threat. The only way to impersonate the server would be to obtain a copy of the private key on the server. At this point we're just simply compromised.

mrwacky42:
Ok, these changes don't hurt anything either..

ariddell:
My sense is that some of these changes *are* unhelpful. They add unnecessary complexity. More importantly, the first and second tips in the linked blog post are connected to the centralized/ICANN PKI infrastructure that many people are trying to move away from.

The ``tls-min-version`` is probably a good idea. Want to put that in a PR?

PR #565, opened 2016-05-27
VPN - enforce TLS min version on the client

mrwacky42:
Enforce `tls-min-version` on the client side configuration.

See also #561


ariddell:
Thanks for this. The target branch should be jessie. (I think requiring TLS 1.2 will break older openvpn clients that are currently working fine.)

mrwacky42:
I'm actually running on CentOS...

mrwacky42:
Closed in favor of #566

PR #567, opened 2016-05-30
Eliminate need for access_compat module in Apache

mikeashley:
The directives provided by `mod_access_compat` are deprecated.  This
patch eliminates references to them.

### DIFF
--- a/roles/git/templates/etc_apache2_sites-available_cgit.j2
+++ b/roles/git/templates/etc_apache2_sites-available_cgit.j2
@@ -12,8 +12,7 @@
     <Directory "/var/www/htdocs/cgit/">
         AllowOverride None
         Options +ExecCGI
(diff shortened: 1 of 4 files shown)


roles/git/templates/etc_apache2_sites-available_cgit.j2
```diff
@@ -12,8 +12,7 @@
     <Directory "/var/www/htdocs/cgit/">
         AllowOverride None
         Options +ExecCGI
-        Order allow,deny
-        Allow from all
+	Require all granted
```
ariddell: Is the spacing off here?

mikeashley: Yes, thank you.  Tab versus spaces.  Corrected.

PR #573, opened 2016-07-06
lint cleanup and one issue fix

oskapt:
Most of these are cleanup after running `ansible-lint` on the master branch. There were a bunch of more-complicated/less-valuable suggestions, but I took on the easy ones. I also fixed stuff that generates deprecation warnings in Ansible >= 2.0.

In doing this I stumbled across a bug in `newebe.conf` where it wasn't notifying supervisor correctly. This has also been fixed (no issue seems to exist for it).

I'll be sending you two more pull requests in coming days that will be built on top of these changes. I'm telling you this now so that you can consider what's coming in your decision to merge the changes provided here. 

One adds a bunch of `no_X.yml` tasks so that users who install X can later uninstall it and have all of the configuration reverted. For example, I run my own bouncer elsewhere, so I don't want `znc` left floating around. I run Openfire for XMPP, so I don't want `prosody` hanging around. I simply replace `znc.yml` with `no_znc.yml` in the `include` directive in `main.yml` for the role, and the magic happens.

The second PR reflects changes that move security items into Ansible Vault. This includes all passwords, certificates, and the tarsnap key. I've updated the documentation accordingly. 

I need to test that those changes work on a default install, which is why they're not included here. I also figure that even if you reject the `no_X` and Vault changes, having some cleanup from `ansible-lint` can't hurt.

Hopefully you'll accept them all. If not, I'll just continue running my own fork of this project.


oskapt:
there's a typo in `ansible.cfg` - inventory location should be `./hosts` instead of just `hosts`. it's fixed in my master branch.

oskapt:
looks like travis is borked.

jplock:
Thanks for these changes.  I think they look pretty good and save to apply.

ariddell:
This is great. Is this all ready? Also -- it would be great to have a separate PR making all these changes targeting the jessie branch.

mikeashley:

mikeashley:
What version of ansible and ansible-lint were you running against when you made these changes?  Your note suggests a version of Ansible > 2.0, but we currently pin master at version 1.6.6, and I didn't see a change to `requirements.txt` or `test-requirements.txt`.

oskapt:
Pinned to Ansible 1.6, eh? No, thanks.

PR #574, opened 2016-07-08
Private variables, service teardown

oskapt:
Sorry if this is messy - I'm doing this for myself and don't seem to be very good at keeping it organized for you.  I'm doing this in its own branch so that new commits don't sneak in after submitting the PR. I'll keep my master pulling from yours so it's clean.

This PR contains everything from the first PR and also:
* private information now stored in `vars/private.yml` (password: `sovereign`)
* documentation updates (new README in `vars` and updates to main README)
* teardown plays for everything except roundcube and mailserver (`no_foo.yml` for each role and task)
* minor updates, like `cache_valid_time` to keep things snappy on repeat runs

I built a default install from the earlier PR and then individually tore down each component to make sure they all work. 

This is the last major PR I'll be submitting for the moment. This branch will be the branch from which I deploy our production server.


oskapt:
in this case travis gets stuck on the vault password. you can solve this by writing out the password (sovereign) to a file and using `--vault_password_file` to point to it in the command.

oskapt:
Added some changes to fix errors discovered today. I tested with installing roles and then uninstalling them, but I didn't test with never installing the roles to begin with. There were some errors on stopping nonexistent services - these are now ignored.

jplock:
These are great PR's, but they are huge.  :)  It would probably help the review and merging process if the various things you are trying to do were broken down into separate PRs that didn't touch so many files at one time.

oskapt:
Is that really appropriate? I'm adding a `no_thing.yml` for every role. It's really the _same_ thing, just in all the roles. Should I submit one PR per role that does the same thing?

Same with moving variables into `private.yml` with Vault. Same thing with fixing all the lint stuff.  Do you want me to submit 50 PRs, one for each part of the same task?

oskapt:
Hate that the close button is closer than the comment button.

mikeashley:

oskapt:
I'm going off on my own.

PR #576, opened 2016-07-13
changed mailserver role to not clobber users/passwords on every run

oskapt:
Sovereign, as written, drops all tables and recreates them on every run. In doing so it recreates all users with passwords as originally set. There is no provision to allow users to change their own passwords (though I'm about to work on that), and if they did, the passwords would be clobbered on the next run. This also creates a race condition - email that arrives during the instant that the database is down will be dropped or misrouted.

This PR changes that. It adds some additional plays to determine what users and domains exist and modifies the template to not add those that do. It also adds an option for `active: False` for each object (domain, user, alias), which will remove it from the database if it exists. Domains are sent through two tests - they must be in a list of domains that exist in the database, and they must not have been deleted during this run. Users and aliases are only added for domains that pass both tests. Finally, in order to keep the `CREATE TABLE IF NOT EXISTS` logic, the three indices were moved to the table create blocks.

The result is that `import.sql` will have only the statements necessary to create the tables and add/remove objects that _need_ to be added or removed. Users or aliases left behind for deleted domains are ignored. Nothing is dropped, so once users have the ability to change their passwords, the integrity of those passwords is preserved, and I, as the admin, do not know what they are. This final point is very important for trust.

This PR is applied against your current master branch, so it is not dependent on anything I've already submitted. I had to provide `ansible.cfg` in order to enable the Jinja `do` extension, so I tossed in some other harmless configuration options that make for nicer command-lines at runtime.

### DIFF
--- a/roles/mailserver/templates/mailserver.sql.j2
+++ b/roles/mailserver/templates/mailserver.sql.j2
@@ -17,11 +17,39 @@ CREATE TABLE IF NOT EXISTS "virtual_users" (
 CREATE TABLE IF NOT EXISTS "virtual_aliases" (
         "id" SERIAL,
         "domain_id" int NOT NULL,
-        "source" TEXT NOT NULL UNIQUE,
+        "source" TEXT NOT NULL,
         "destination" TEXT NOT NULL,
         PRIMARY KEY ("id"),
         FOREIGN KEY (domain_id) REFERENCES virtual_domains(id) ON DELETE CASCADE
 );
+
+{# this entire block is from [here](http://dba.stackexchange.com/questions/35616/create-index-if-it-does-not-exist) #}
+{% if postgres_minor.stdout|int <= 3 %}
+{# 9.3 and lower #}
+DO $$
+BEGIN
+
+IF NOT EXISTS (
+    SELECT 1
+    FROM   pg_class c
+    JOIN   pg_namespace n ON n.oid = c.relnamespace
+    WHERE  c.relname = 'source_idx'
+    AND    n.nspname = 'public' -- 'public' by default
+    ) THEN
+
+    CREATE INDEX source_idx ON public.virtual_aliases (source);
+END IF;
+
+END$$;
+{% elif postgres_minor.stdout|int == 4 %}
+IF NOT EXISTS (
+  SELECT to_regclass('public.source_idx');
+  ) THEN
+  CREATE INDEX source_idx ON public.virtual_aliases (source);
+{% elif postgres_minor.stdout|int > 4 %}
+CREATE INDEX IF NOT EXISTS source_idx ON public.virtual_aliases (source);
+{% endif %}
+
 
 {# set a place to capture deleted domains #}
 {% set deleted_domains = [] %}
@@ -115,7 +143,7 @@ INSERT INTO "virtual_aliases" ("domain_id", "source", "destination")
 (not virtual_alias.active|default(True) or
 source_domain in deleted_domains or
 dest_domain in deleted_domains) or
-virtual_alias.destination not in mail_users.stdout_lines %}
+virtual_alias.destination in deleted_users %}
 {# inactive alias or domains deleted, but is in database - remove it #}
 DELETE FROM "virtual_aliases" WHERE
     domain_id = '{{ virtual_alias.domain_pk_id }}' AND
--- a/roles/mailserver/templates/mailserver.sql.j2
+++ b/roles/mailserver/templates/mailserver.sql.j2
@@ -1,16 +1,9 @@
--- If tables are not dropped, have to truncate before insert or use "insert or replace" (not postgres compatible)
-
-DROP TABLE IF EXISTS "virtual_users";
-DROP TABLE IF EXISTS "virtual_aliases";
-DROP TABLE IF EXISTS "virtual_domains";
 
 CREATE TABLE IF NOT EXISTS "virtual_domains" (
         "id" SERIAL,
-        "name" TEXT NOT NULL,
+        "name" TEXT NOT NULL UNIQUE,
         PRIMARY KEY ("id")
 );
-
-CREATE UNIQUE INDEX name_idx ON virtual_domains (name);
 
 CREATE TABLE IF NOT EXISTS "virtual_users" (
         "id" SERIAL,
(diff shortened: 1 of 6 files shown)


roles/mailserver/templates/mailserver.sql.j2
```diff
@@ -21,37 +14,88 @@ CREATE TABLE IF NOT EXISTS "virtual_users" (
...
 );
 
-
-CREATE UNIQUE INDEX email_idx ON virtual_users (email);
-
 CREATE TABLE IF NOT EXISTS "virtual_aliases" (
         "id" SERIAL,
         "domain_id" int NOT NULL,
-        "source" TEXT NOT NULL,
+        "source" TEXT NOT NULL UNIQUE,
```
carljm: This is wrong -- if you want a fan-out alias, you can specify the same source multiple times with different destinations. I make use of this, please don't break it. If you want a unique logical id for this table, you need to use source and destination together.

carljm:
In general I'm in favor of this approach (especially due to the race condition issue, though I wonder if a transaction and appropriate transaction isolation could solve that too?) But the trust issue isn't necessarily a problem with the current approach, you can provide a separate UI for users to provide a password and you just get the hash to integrate into the Ansible config. You're going to have access to their password hash regardless, that's not a problem.

roles/mailserver/templates/mailserver.sql.j2
oskapt: fek - my bad. it wasn't supposed to be `UNIQUE` like the others. i'll fix now.

oskapt:
Realized after posting PR that it didn't handle a case where a user was removed from the configuration but still in the db. I now check for that and remove from the db (config is god). Same for aliases - if no user exists (deleted in this run or otherwise), aliases are removed.

oskapt:
@carljm - My thought here is that users changing passwords updates the database (if I use the roundcube plugin, which is what I'll be poking at next). If passwords are already in the database, then I just need to not clobber them. That's a _lot_ less work than having some external setup that grabs the password, encrypts it, puts it in Ansible, and then runs Ansible to update the database. Users also expect their password change to be instantaneous. Even if I update the database directly, then extract those passwords and update Ansible, it's clunky. Now my mailserver is updating my sovereign repository, and for most users, the Ansible configuration doesn't live on the mailserver. Git, SSH, push permissions...what I have here is far easier to manage.

carljm:
@oskapt yeah, I get your thinking, and I've wrestled back and forth with this same question. But it still seems to me that there's some conceptual inconsistency in this PR. Right now, as you say, "config is god," in every aspect. Whatever's in the config is guaranteed to match what's actually used (assuming you've run Ansible.)

After this PR, "config is kinda-sorta-mostly-god, but only in terms of which records exist, not necessarily in terms of their data attributes -- various things could be changed directly in the database and will then not match your config, and that inconsistency could persist indefinitely." I understand why that seems useful for passwords, but it definitely doesn't seem like an improvement in terms of simplicity or clarity of the mental model.

I wonder if there'd be some way to do this that applied _only_ to password hashes, and removed password hashes from the Ansible config entirely? I just really don't like the idea of having stuff in Ansible that is permanently out-of-sync and misleading.

oskapt:
@carljm It's a dilemma, for sure. I've run sovereign for ~2y at home, where my wife doesn't care if she can't change her password. I'm deploying it for a small office environment now, and they _do_ care. The workflow now with Fastmail is I set them up and give them their initial password and then tell them to change it. If they forget it, I can go back and reset it. I see this change for sovereign as the same logic. I set your initial password in the config and you change it. From that point on the password hash is ignored, as I believe _it should be_. I could add a `force: True` flag that would act as a password reset, so that the "setting and resetting of passwords" is still within the sysadmin domain, whether it's Ansible or some other tool.

In this configuration there is no other tool, so enabling Ansible to act like any other admin/management tool for passwords seems like the right move.

I don't know whose decision it is to merge or not merge these things - I'm just handing them over because they seem useful. I'm running them all in production, and they all solve specific problems I've encountered w/ sovereign.

FYI - I fixed the index issue with virtual aliases. I've only tested it on Postgres 9.1. If someone can test the IF/THEN for 9.4, that would be great. Otherwise w/ 9.5 it solves itself.

carljm:
@oskapt I'm not a maintainer/committer, so the merge decision isn't up to me :-) Definitely appreciate you sharing your work back!

ariddell:
I appreciate how this solves a real problem. It does, however, as @cantsin points out, run contrary to the idea of the ansible configuration specifying precisely the desired state of the server.

oskapt:
I think the answer to "should I know every user's password and should they never be able to change them" trumps the philosophical dilemma of "should Ansible control everything."  Adding `force: true` or something still gives me the ability to override it during Ansible.

Ansible replaces actions that I would do manually, assuring that they are done consistently. Any action that I would do, it does for me. I would reset a password when it's lost. I would not reset everyone's password every time I add a new account.

mikeashley:
@oskapt I appreciate the effort you put into this.  Thank you.  Could you separate the problem(s) you are trying to solve from your implementation?  @carljm is making some good points, for example, about alternative and probably more robust ways of addressing the race condition that exists when rerunning the mailserver role concurrent with mail delivery.  As for password management, I want to be sure I understand what you're trying to accomplish.  I suggest opening two separate issues, and let's work through it there so we can document the thinking.

oskapt:
Please see #581 and #582 for the issues you requested.

I move fast. Sorry that I don't open issues first. I'm the sole admin in a fast-moving company with little tolerance for stuff not working. They don't understand much about IT, so I don't have the luxury of working through solution options over several weeks - when I find something that doesn't work, I fix it. If I don't, I have to answer impossible questions from the company owner.

FWIW, I've been building infrastructure on the Internet for more than 20 years, and was the owner and principal architect for an MSP that built and managed datacenters for US media companies for 14 years. That's the foundation for the solutions I'm proposing.

All of the solutions you get are also post-production. I'm running them long before I submit them back to you for inclusion. That's why they're often difficult to extract into neat little branches. I actually reconstructed my whole repo yesterday so that I can give you cleaner branches to pull from while still keeping my `master` branch as the collection of all the stuff I'm actually using. I can't currently use the `master` branch because it's hung up on #573 and #574. 

If you want, you can close all of them, and we can work through them in the issues. I can generate  new PRs if you decide that you want a solution that uses the configuration I've written. They're all currently failing Travis anyway, either because they're using Ansible Vault or for some other Travis reason that isn't connected to the code I've written.

mikeashley:

oskapt:
I'm going off on my own.

PR #578, opened 2016-08-04
make letsencrypt output silent (only output on error)

Dosenpfand:
Automatic renewal of certificates should occur silently if there was no error, currently a daily reminder is sent by mail that nothing happened.
This is also the [official recommendation](https://certbot.eff.org/docs/using.html#renewal):
>If you’re sure that this command executes successfully without human intervention, you can add the command to crontab (since certificates are only renewed when they’re determined to be near expiry, the command can run on a regular basis, like every week or every day). **In that case, you are likely to want to use the -q or --quiet quiet flag to silence all output except errors.**


mikeashley:
Thank you!

PR #580, opened 2016-08-09
Roundcube from source

oskapt:
I'm going to get flamed by the "don't break Debian" crowd, but whatevz. A mailserver with no webmail client is not nearly as useful as a mailserver with one. Ubuntu 14.04 is also stuck on 0.9.5.

This PR installs Roundcube 1.2.1 from (**_gasp_**) SOURCE!

OMG. Unreal. A bunch of PHP scripts in a docroot that we don't need a package maintainer for! Madness! Cats and dogs living together! It's like 1996 again! Let's install Slackware!

This includes a little filter plugin that I needed to get the plugins wrapped in quotes in the template in some fashion that didn't make me want to claw my eyes out. It's called 'wrap' and was not my idea. It came from something on stackoverflow or someplace...DRY, etc.

I'm running this in production, and I (painfully) extracted the core of it today to make this PR. I ran a basic install with default variables and it didn't scream, so it should be good. It's better with my other PRs (password plugin, etc), but those aren't necessary...however, this PR includes previous changes to install the roundcube password plugin because I didn't want to deal w/ manually merging the two branches. I compensated by not activating it - it's just code sitting there, waiting for the user to find it (or for my other PR to be merged).

I _did_ have to hack their db init script so that it can be idempotent and safely run without barfing on every Ansible run.

This may be useful for work being done in #495 and #507.

### DIFF
--- a/filter_plugins/core.py
+++ b/filter_plugins/core.py
@@ -1,0 +1,8 @@
+def wrap(list):
+    return [ '"' + x + '"' for x in list]
+
+class FilterModule(object):
+    def filters(self):
+        return {
+            'wrap': wrap
+        }
--- a/roles/webmail/files/roundcube_plugins_managesieve_config.inc.php
+++ b/roles/webmail/files/roundcube_plugins_managesieve_config.inc.php
@@ -1,0 +1,100 @@
+<?php
+
+// managesieve server port. When empty the port will be determined automatically
+// using getservbyname() function, with 4190 as a fallback.
+$config['managesieve_port'] = null;
+
+// managesieve server address, default is localhost.
+// Replacement variables supported in host name:
+// %h - user's IMAP hostname
+// %n - http hostname ($_SERVER['SERVER_NAME'])
+// %d - domain (http hostname without the first part)
+// For example %n = mail.domain.tld, %d = domain.tld
+$config['managesieve_host'] = 'localhost';
+
+// authentication method. Can be CRAM-MD5, DIGEST-MD5, PLAIN, LOGIN, EXTERNAL
+// or none. Optional, defaults to best method supported by server.
+$config['managesieve_auth_type'] = null;
+
+// Optional managesieve authentication identifier to be used as authorization proxy.
+// Authenticate as a different user but act on behalf of the logged in user.
+// Works with PLAIN and DIGEST-MD5 auth.
+$config['managesieve_auth_cid'] = null;
+
+// Optional managesieve authentication password to be used for imap_auth_cid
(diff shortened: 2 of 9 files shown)


mikeashley:
This is essentially a duplicate of PR #457 which we elected not to take.  Please consolidate discussion to #507.  We do need to get the webmail situation resolved.

neuhaus:
We've had so many pains with the distribution provided packages, perhaps for roundcube, building from source is the better option? I'm open to this idea (as I mentioned in the closing comment of #457 ).

oskapt:
The argument for using a package is that it allows for easy uninstalling of the content if you remove it. That barely applies here - Roundcube isn't throwing stuff in `/usr/local/{etc,bin,sbin,share}/` and `/var/{lib,log,spool}/` or whatever. It's a web app. It's PHP.

I'm putting it all in `/var/lib/roundcubemail-{{ roundcube_version }}` with a link to that at `/var/lib/roundcube`. All config files have been updated/modified to use this path. My `no_roundcube.yml` from #574 removes these locations, and all configuration, whether it was installed by package or from source.

We have some basic choices:
- use 0.9.5 (not an option on jessie)
- use nothing 
- install it from source
- wait for someone upstream to fix whatever the problem is

Given these options, let's consider our objective. If we want to be package maintainers for Debian or Roundcube, then we're in the wrong place. If we want to be proud that our collection of tools for running a private mailserver does _not_ provide a web mail client, then maybe we're doing alright. If we want to wait for Debian to change their mind about including it, or for someone else to write another web mail client that Debian _will_ include, then we should all go get beers.

If, however, we want to give a complete tool set for privatization of mail in the post-Snowden world where we all live, we can consider the shortest path to the objective of giving our users a functional, robust, web mail client and install it from source.

oskapt:
I know y'all will prob reject this PR, but that doesn't stop me from updating the branch to make sure it works.

Found today that if this is installed on top of a packaged install of roundcube, it barfs. I added some plays to remove the packaged versions and check for/optionally delete `/var/lib/roundcube` so that we can symlink it.

I have this running in production on 14.04.

mikeashley:

mikeashley:
There seem to be no takers for this PR.  At a minimum, this PR needs to be reworked to install Roundcube cleanly and not bother with uninstalling an existing package.  The role should assume a clean install of Sovereign.  I recommend looking at the commits merged at commit 0e379f as an example of how to do this.

Please do resubmit a reworked PR if you care to do so.  I'm convinced now that we're going to need to install from source to get Roundcube set up properly.

PR #588, opened 2016-08-30
Allow SSH agent forwarding when using sudo

tomasbedrich:


mikeashley:
Why would we want this change?

tomasbedrich:
It allows you to use SSH agent forwarding from local machine when logged in as a root on the remote machine. I use this feature for cloning the Github repositories to /var/www (owned by root) without a need to store my private SSH keys on the remote machine.

mikeashley:
Thank you!

PR #589, opened 2016-08-30
Update Rspamd setup

tomasbedrich:
Rspamd changed the recommended way of starting its services, thus the playbook is failing on this task.

References:
- https://github.com/vstakhov/rspamd/pull/762
- https://rspamd.com/doc/quickstart.html#running-rspamd

> Packaging should start rspamd and configure it to run on startup on installation.


carljm:
FWIW, I wrote this rspamd setup initially, and this change looks correct to me.

mikeashley:
Thanks, guys!

PR #590, opened 2016-09-05
Reinstate webmail role (take 2)

mikeashley:
This PR reinstates the webmail role lost when we moved to Jessie.  It installs Roundcube from source using sqlite3 as the database backend.  The `carddav`, `managesieve`, and `twofactor_gauthentication` plugins are carried forward using the plugin management system that the Roundcube team has implemented using PHP composer.

### DIFF
--- a/roles/webmail/templates/var_www_roundcube_config_config.inc.j2
+++ b/roles/webmail/templates/var_www_roundcube_config_config.inc.j2
@@ -1,0 +1,86 @@
+<?php
+
+/*
+ +-----------------------------------------------------------------------+
+ | Local configuration for the Roundcube Webmail installation.           |
+ |                                                                       |
+ | This is a sample configuration file only containing the minimum       |
+ | setup required for a functional installation. Copy more options       |
+ | from defaults.inc.php to this file to override the defaults.          |
+ |                                                                       |
+ | This file is part of the Roundcube Webmail client                     |
+ | Copyright (C) 2005-2013, The Roundcube Dev Team                       |
+ |                                                                       |
+ | Licensed under the GNU General Public License version 3 or            |
+ | any later version with exceptions for skins & plugins.                |
+ | See the README file for a full license statement.                     |
+ +-----------------------------------------------------------------------+
+*/
+
+$config = array();
+
+// Database connection string (DSN) for read+write operations
+// Format (compatible with PEAR MDB2): db_provider://user:password@host/database
+// Currently supported db_providers: mysql, pgsql, sqlite, mssql, sqlsrv, oracle
+// For examples see http://pear.php.net/manual/en/package.database.mdb2.intro-dsn.php
+// NOTE: for SQLite use absolute path (Linux): 'sqlite:////full/path/to/sqlite.db?mode=0646'
+//       or (Windows): 'sqlite:///C:/full/path/to/sqlite.db'
+$config['db_dsnw'] = 'mysql://roundcube:pass@localhost/roundcubemail';
+
+// The mail host chosen to perform the log-in.
+// Leave blank to show a textbox at login, give a list of hosts
+// to display a pulldown menu or set one host as string.
+// To use SSL/TLS connection, enter hostname with prefix ssl:// or tls://
+// Supported replacement variables:
+// %n - hostname ($_SERVER['SERVER_NAME'])
+// %t - hostname without the first part
+// %d - domain (http hostname $_SERVER['HTTP_HOST'] without the first part)
+// %s - domain name after the '@' from e-mail address provided at login screen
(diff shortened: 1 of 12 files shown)


neuhaus:
Thanks, looks good. Do you have it running on your test system?

mikeashley:
Yes.

neuffer:
On 05.09.2016 12:42, Mike Ashley wrote:
>
> This PR reinstates the webmail role lost when we moved to Jessie. It
> installs Roundcube from source using sqlite3 as the database backend.
> The |carddav|, |managesieve|, and |twofactor_gauthentication| plugins
> are carried forward using the plugin management system that the
> Roundcube team has implemented using PHP composer.
>

Why do you use sqlite3 when there is already an installed postgresql?
This makes life just much harder if you want to integrate the system a
bit more.

Cheers
  Mike

mikeashley:
What do you mean by "integrate the system"?  Roundcube does not persist much data, and I think it's an advantage to not have this role depend on postgresql.

roles/webmail/templates/var_www_roundcube_config_config.inc.j2
```diff
@@ -25,7 +25,7 @@ $config = array();
 // For examples see http://pear.php.net/manual/en/package.database.mdb2.intro-dsn.php
 // NOTE: for SQLite use absolute path (Linux): 'sqlite:////full/path/to/sqlite.db?mode=0646'
 //       or (Windows): 'sqlite:///C:/full/path/to/sqlite.db'
-$config['db_dsnw'] = 'mysql://roundcube:pass@localhost/roundcubemail';
+$config['db_dsnw'] = 'sqlite:////var/www/roundcube/db/sqlite.db?mode=0664';
```
mikeashley: Locating the database file here means it doesn't get backed up.  This should be moved to a location such as `/decrypted/roundcube/sqlite.db`.

neuffer:
On 05.09.2016 15:49, Mike Ashley wrote:
>
> What do you mean by "integrate the system"? Roundcube does not persist
> much data, and I think it's an advantage to not have this role depend
> on postgresql.
>
We already have the dependency on postgresql, so you don't gain
anything, but instead complicate the dependency graph and in case of
upgrades
you now have to migrate 2 database backends and their data instead of one.

I am planning to add postfixadmin to the mix, which will also use
postgresql as its backend and to change the current postfix setup
accordingly
Keeping all data in one database server instead of distributing it
amongst postgresql, the mysql family and sqllite gives you the advantages
mentioned above plus it is easier for those pieces of software to
interact with each other.

Somewhere on the net there is a postfixadmin/roundcube integration
plugin floating around  which will then allow end users to change their
passwords from the roundcube gui for example.

Keeping you data in several seperate and different database backends
will turn this into a mess.

Cheers
  Mike

mikeashley:
Mike, sorry, I don't follow your reasoning.  On an upgrade, you have to migrate the database no matter where it is.  As for interaction, that typically happens through the application layer of the packages we're talking about, not the persistence layer, so no advantage there either.

You are right we are adding a package dependency (sqlite3), but in exchange we get a more efficient runtime.  That's a benefit: we're aiming to keep our memory footprint small.  Our goal is to get rid of mysql (see Issue #502), and I think it'd be great to eliminate postgres also.

neuffer:
On 07.09.2016 01:43, Mike Ashley wrote:
>
> Mike, sorry, I don't follow your reasoning. On an upgrade, you have to
> migrate the database no matter where it is. As for interaction, that
> typically happens through the application layer of the packages we're
> talking about, not the persistence layer, so no advantage there either.
>

And how many applications do you know support accessing more then one
database engine at one? It is always either or.
How are you doing to keep the account information for dovecot, and
roundcube and... in sync when they are all kept in different DB engines?
Either you are building yourself a single user mail server and you don't
care or you'll get into into support hell real quick.
And then you'll find out that maintaining accounts and passwords accross
multiple applications AND database engines is no fun.
It is much easier if you have to access one, to code for one instead of
two, three or more that you access in parallel.
How many persistence layers do you know support different engines in
parallel? You always configure them either or.

> You are right we are adding a package dependency (sqlite3), but in
> exchange we get a more efficient runtime. That's a benefit: we're
> aiming to keep our memory footprint small. Our goal is to get rid of
> mysql (see Issue #502
> <https://github.com/sovereign/sovereign/issues/502>),
>

Then standardize on postgresql as described there.

> and I think it'd be great to eliminate postgres also.
>
With sqlite you run into almost the same problem as descriped above, one
application one engine. You'll run multiple engines and your data is
separate.

In that case sovereign won't my kind of system to support anymore and
I'll go back to hand crafting my installation scripts.

Issue #557, opened 2016-05-07
OpenDMARC cron failing


tomasbedrich:
I am using Debian Jessie and everything is working fine, except of this cron task, which is failing. Here is the log:

```
opendmarc-import: started at Fri May  6 01:00:02 2016
opendmarc-import: connected to database
opendmarc-import: terminating at Fri May  6 01:00:02 2016
opendmarc-reports: started at Fri May  6 01:00:03 2016
opendmarc-reports: failed to collect domain names: Table 'opendmarc.requests' doesn't exist
opendmarc-import: started at Sat May  7 01:00:01 2016
opendmarc-import: connected to database
opendmarc-import: terminating at Sat May  7 01:00:01 2016
opendmarc-reports: started at Sat May  7 01:00:02 2016
opendmarc-reports: failed to collect domain names: Table 'opendmarc.requests' doesn't exist

**Comment by mikeashley** [comment]
I am not seeing that on my server.  Perhaps replay the mailserver playbook with the dmarc tag and see if there is anything strange?  From the log it looks like the database is misconfigured.

**Comment by tomasbedrich** [comment]
Replaying the playbook with `--tag=dmarc` completed successfully and gave me this result:

```
...
PLAY RECAP *********************************************************************
ok=31   changed=0    unreachable=0    failed=0
```

Nothing was changed and everything seems to be okay from the perspective of playbook.

I was digging a little bit deeper and opened the MySQL DB and *it was empty* - so even after replaying the playbook, the schema hasn't been created. Therefore I have created the schema manually using `mysql -u opendmarc -p < /usr/share/doc/opendmarc/schema.mysql` and it works now.

The DB was empty - my fault, I know. But I think there is also a problem with the playbook, more concretely [with this line](https://github.com/sovereign/sovereign/blob/jessie/roles/mailserver/tasks/opendmarc.yml#L29). The notify handler won't execute unless DB is created so *the schema cannot be re-created separately*. What do you thing about separating the schema import task?

carljm:
I made that change. Previously the schema always re-imported on every run. 

Ideally the schema import would be a task that runs conditionally based on a more detailed check (like, does the database exist, or do the tables exist). The ideal check would be for a complete schema match, not sure that's feasible.

Anyway if you're able to put together a PR along those lines I'd be in favor, but I don't think we should go back to "schema import always runs unconditionally."

carljm:
Probably what happened in your case was that an Ansible run crashed after creating the db but before handlers were run.

[...]

PR #591, opened 2016-09-16
Make opendmarc database setup more robust

mikeashley:
This PR addresses Issue #557.  I tested it by deploying, redeploying to verify no changes, manually dropping the database, and deploying a final time to confirm the database is configured again.


ariddell:
Looks like a simple, sensible change.

PR #592, opened 2016-09-26
Added abuse as a default alias

bpeterman:
https://pingability.com/zoneinfo.jsp complains when mail is rejected for the abuse user.


neuhaus:
Thanks!

PR #593, opened 2016-09-27
Setup postgres in common role

wsteitz:
Postgres is used by several roles, but the setup is currently part of the 'mailserver' role. By moving it to 'common', it's possible to disable the mailserver without breaking the others.


neuhaus:
Looks good to me, have you tested this?

wsteitz:
I tried it on a newly setup VPS without problems.

neuhaus:
Thanks!

PR #598, opened 2016-10-08
Use wallabag v2

wsteitz:
This PR upgrades wallabag to the latest version. Wallabag 2 was released in April 2016 and brings many improvements and changes. Looks like they re-wrote many parts of the application. The old 1.x version is not actively worked on anymore.

Two things that still need some work (help / suggestions welcome!)
1. currently, the URL is read.myhostname/web. I think the apache config files needs some tweaking. Any ideas?
2. re-runing the "Install wallabag" step results in an error.


mikeashley:
Thank you for the PR.  Would it be possible to split the configuration file changes into different commits?  I find it very helpful to have one commit for the package's configuration file and then a second commit for sovereign-specific changes.  This makes it easier in the future to reason about configuration file changes in new versions of the package.

wsteitz:
I've splitted the commits in role and configuration changes. Of course that means that the intermediate commit is not a working revision, as the steps in the role don't match the configuration files.

Re-running the installer now also works without problems. To avoid overwriting the old wallabag 1.x database, the database is now called wallabag2.

Overall, the PR is now feature complete and ready for review.

mikeashley:
Thanks for the updates.  Two other questions after review: 1) how have you tested this, and 2) have you considered adding `DESIGN.md` to document assumptions, intent, gaps, etc?  You might look at `DESIGN.md` in the webmail role for inspiration.

roles/readlater/tasks/wallabag.yml
```diff
@@ -61,21 +64,20 @@
...
-            group=www-data
+    - bin
+    - app/config
+    - vendor
+    - data
+    - app
+    - var
 
 - name: Rename existing Apache wallabag virtualhost
   command: mv /etc/apache2/sites-available/wallabag /etc/apache2/sites-available/wallabag.conf removes=/etc/apache2/sites-available/wallabag
```
wsteitz: That's a leftover from the prior implementation, I'll check if it save to remove...

Thanks for your explanation on the upgrade procedure of sovereign! Good to know that backwards compatibility is not necessary.

roles/readlater/tasks/wallabag.yml
```diff
@@ -49,10 +34,28 @@
...
+  command: php /root/composer.phar install --no-dev -o --prefer-dist
            chdir=/var/www/wallabag
            creates=/var/www/wallabag/vendor/autoload.php
+  environment:
+    SYMFONY_ENV: prod
+
+- name: Install wallabag
+  command: php bin/console wallabag:install --env=prod
+           chdir=/var/www/wallabag
+           creates=/var/www/wallabag/bin/doctrine.php
```
wsteitz: Ok, will do.

roles/readlater/templates/var_www_wallabag_app_config_parameters.yml.j2
```diff
@@ -0,0 +1,29 @@
+parameters:
+    database_driver: pdo_pgsql
+    database_host: localhost
+    database_port: ~
+    database_name: {{ wallabag_db_database }}
```
wsteitz: Now I get what you are after! I think there is no default config in the wallabag source code, as it is generated during installation time. But I'll find out the default values

akovalyov:
is there any chance to get it merged soon?

akovalyov:
@mikeashley @wsteitz how can I help in getting this into core? I am a newbie in Ansible, but at least can help with testing (not with documentation, English and writing are two my weak points as well).

wsteitz:
@akovalyov  I decided against using ansible on my server. So there is no real interest from my side to finish up on this one. Sorry.

In the meantime, wallabag's installation steps got simplified a bit. It would need to check what has changed and adapt accordingly. the best idea is probably closing this PR and re-open a new one.

Issue #601, opened 2016-10-15
Multiple configuration errors


working-name:
Hello guys,

I'm about to give up. This has been an exercise in futility overall.

I tried both master and wheezy on 14.04, 15.04, 16.04, useless. No way in hell you can get it to work, there's just too many things that are different, packages that are discontinued, owncloud has issues with requirements not updating, there's a test repo to address that instead of actually updating the official repo, whatever. No matter how many times I tried to run ansible-playbook it kept failing in new ways, which utterly defeats the purpose of automating commands.
### changes I made while attempting to get 'wheezy' to complete on 16.04

``` py
roles/common/tasks/encfs.yml
13:# TODO: "Add mail user to fuse group" > did you forget to create the group "fuse"? Is it to be created by fuse itself? Didn't quite work if that's the case.

roles/common/tasks/google_auth.yml
7:    #TODO: "Ensure required packages are installed" > libpam-google-authenticator package is available in 16.04 now, re-enabled it
12:# TODO: "(?:Download|Extract|Install) Google authenticator" commented out. GoogleCode is dead.
41:#TODO: "Generate a timed-based, no reuse, rate-limited (3 logins per 30 seconds) with one concurrently valid code for default user" > path is not /usr/local/bin/google-authenticator for 16.04, it is /usr/bin/google-authenticator

roles/common/tasks/main.yml
72:# TODO: "Set ServerName for Apache" change 'dest=/etc/apache2/conf.d/fqdn' to  'dest=/etc/apache2/conf-available/fqdn.conf'

roles/git/tasks/cgit.yml
24:#TODO: "Build and install cgit" fails because googlecode.com is deadm and at some point it tries to pull it from there.

roles/git/tasks/gitolite.yml
34:#TODO: "Copy SSH public key to server" refers to gitolite.pub but the file is not in the git repo at all. Does one need to create it? 'ssh-keygen -t rsa -f roles/git/files/gitolite' if so.

roles/git/tasks/main.yml
5:# - include: cgit.yml tags=cgit #TODO: disabled "- include: cgit.yml tags=cgit" because it fails to compile, wants some googlecode thing.

roles/mailserver/files/etc_opendmarc_import.sql
5:-- TODO: lastsent TIMESTAMP NOT NULL DEFAULT '2016-01-01 12:00:00', => the default value was causing an error on import with postgresql 9.5+173 on xenial

roles/mailserver/tasks/dovecot.yml
[...]

PR #608, opened 2016-10-29
Correct box for debian jessie

Titan-C:
I wanted to debug why I don't get my vps to work. I have problems as listed in https://github.com/sovereign/sovereign/issues/601#issuecomment-255566608

I followed the wiki for development but it could not find the image. This little fix does the job

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -36,7 +36,7 @@ Vagrant.configure('2') do |config|
 
   # Debian 8 64-bit (officially supported)
   config.vm.define 'jessie', primary: true do |jessie|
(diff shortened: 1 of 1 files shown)


mikeashley:
Thanks for this.  I tried the patch on my machine, and I was not able to download the image.  Browsing at hashicorp, though, it does look correct.  Is a hashicorp login needed now to download these boxes?

Titan-C:
I did not need any login. After changing the image name it worked for me.

mikeashley:
Thanks.  I had collision with Anaconda on my machine.  Looks good.

PR #609, opened 2016-10-30
Make sure /decrypted/roundcube is writable by roundcube/apache/www-data

hippich:


mikeashley:
Thanks.  Did this fix a specific problem for you?

hippich:
Yeah, sorry, I forgot to describe issue I had - in my case /decrypted/roundcube/sqlite.db was missing (not sure why) and roundcube was trying to create it on a first open, but failing due permissions. And with group set to www-data it successfully creates default sqlite.db DB and let me log in.

PR #610, opened 2016-10-30
Force git pull when pulling wallabag

hippich:
In few instances I had situation where certain files were removed from wallabag installation (probably by wallabag itself, or by sovereign) which resulted in process breaking due git refusing to pull updates into repo with uncommited changes. Force=yes should get around it.


mikeashley:
PR #598 also forces the pull.  @hippich Are you interested in using that PR?  It would be nice to get some testing on it before we integrate.

hippich:
Yeah, #598 definitely sounds like a better option. Closing this one. Thanks for pointing out it to me!

Issue #149, opened 2014-01-26
Error at "Set postgres password"


Korni22:
Up-to-date Debian (last update 26.01.14) and up-to-date playbook (cloned on 26.01.14), getting the following error:

```
TASK: [mailserver | Set postgres password] ************************************
failed: [**.**.**.**] => {"changed": true, "cmd": ["sudo", "-u", "postgres", "psql", "-d", "postgres", "-c", "ALTER USER postgres with  password '****';"], "delta": "0:00:00.082260", "end": "2014-01-26 14:41:42.101328", "item": "", "rc": 2, "start": "2014-01-26 14:41:42.019068"}
stderr: perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
    LANGUAGE = "en_US:en",
    LC_ALL = (unset),
    LC_CTYPE = "de_DE.UTF-8",
    LANG = "C"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
psql: could not connect to server: No such file or directory
    Is the server running locally and accepting
    connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

Seems to be my fault though, I am trying to fix this by setting my locale - despite that, it might be good to know for the devs...

PR #613, opened 2016-10-30
Set default locales in order to start postgresql

akovalyov:
Hi,
I had a trouble with installation.
selection_069

It appeared, that postgresql needed correct locale to be exported. This is very simple and naive fix. However, it works.

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -59,6 +59,9 @@
 - name: Set decrypted directory permissions
   file: state=directory path=/decrypted group=mail mode=0775
 
+- name: "Configure locales"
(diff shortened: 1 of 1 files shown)


jplock:
Can this be broken out into separate task statements?

lukecyca:
I see we have a well-hidden wiki page about this from when it came up in #149
https://github.com/sovereign/sovereign/wiki/Troubleshooting

Indeed it would be better to handle it from within ansible, or otherwise make it more user-friendly.  However I have some concerns with this patch:
- forces `en_US.UTF-8` whereas some users may not be english-speaking 'mericans.
- writes out `/etc/default/locale` directly. Should it use `dpkg-reconfigure locales`?
- my reading of this shell script is that it will append two more lines onto `/etc/default/locale` every time the playbook is run!

PR #615, opened 2016-11-04
Add Baïkal role

cimm:
A simple role to install the [Baïkal](http://baikal-server.com) CalDAV (calendars) and CardDAV (contacts) server via Sovereign.

The server runs on `dav.example.com` and the connection is encrypted  thanks to the Let's Encrypt support Sovereign already provides.

### Why

I don't need a full fledged ownCloud installation. I've heard good things about it but I simply don't need file sharing, a web UI for my calendars, sharing, photo galleries, etc. You might, and that's fine, but for me it only adds complexity (and possibly, maybe, a larger attack surface). I only need a something that syncs my contacts and calendars between my phone and laptop. That's all.

### Alternatives

There are a few other CardDAV and CalDAV solutions out there. I personally tried [Radicale](http://radicale.org/) and Baïkal but there are [plenty](https://en.wikipedia.org/wiki/Comparison_of_CalDAV_and_CardDAV_implementations) of other options out there. There is even a Radicale Sovereign role, currently being discussed in pull request #462.

Why not go for Radicale? I tried. I tried again this week but ran an older version a while back and it doesn't fit my bill. I like it, it's closer to the unix philosophy as Baïkal but development doesn't feel as active and there have been issues when using macOS software (Contacts and Calendar). Not saying it's a bad solution, it just didn't work for me.

### Open issues

- [ ] Replace MySQL with PostgeSQL

This role currently depends on MySQL and it's my understanding Sovereign is moving to PostgreSQL. If there is a demand for this role I will gladly update the role to use PostgreSQL.

- [ ] Tests

There are currently no tests for this role. I did not yet take the time to learn how to test an Ansible role and will add tests if the there is intertest in a Baïkal role.

- [ ] How do we include options?

I currently added the Baïkal role to the `site.yml` config. This is not a good idea as you probably don't want to run ownCloud _and_ Baïkal. How do we show there is an alternative without adding it to the `site.yml` config? Add it as a comment at the end or is there a better way?

- [ ] GUI configuration

Once installed Baïkal needs to be configured via a web UI. Once done the web UI can be disabled but you need it the first time. This bothers me as it means the role is not fully automated: you need to configure it manually, a config file would have been better. Is this simply a matter of explaining the process in the README or can we do better?

### Feedback

Feedback welcome: is this something Sovereign can use or should we just focus on the ownCloud integration and not spend time supporting alternative options?


neuhaus:
Have you tried using SQLite instead of MySQL? Is the performance good enough?

Regarding site.yml, I think you should add it but comment it out and add a comment there (or explain it in the relevant section of the README).

GUI configuration: What is being done there? Can we automate it?

No MySQL backup: You're right. That's a bug. I'll open a new issue.

cromulus:
If we remove owncloud, could I suggest [seafile](https://www.seafile.com) as a file sharing alternative? My strong preference is to choose projects that do one thing and do that one thing well..

neuhaus:
I don't see a reason to remove owncloud. We're just offering more than one alternative. Perhaps it's a mistake not to focus, time will tell

ariddell:
@cromulus I agree with you that owncloud does too much. The broader issue is that sovereign (or rather an ansible playbook + roles) is monolithic and it's hard to do anything except remove roles. It's not easy to provide a set of alternatives to choose from or to allow third-party "plugin" roles. Ansible doesn't seem to be the best tool in such a situation though.

neuhaus:
One reason to use alternatives to Owncloud could be performance and resource requirements. Owncloud 9.1 recommends 512MB of memory (the minimum is 128MB).
If you run sovereign on a constrained device like a Raspberry Pi you want to have very lean services to get acceptable performance. On the other hand if you have more resources you may like the extra features that Owncloud offers.

jplock:
I'd prefer to stick with owncloud as I find their desktop and mobile apps to be of decent quality.

cromulus:
@ariddell Hmm... yes. My hunch is that the right long term move for personal mail, cardav, caldav, filesync, etc. etc. is to setup a docker ecosystem that revolves around an ldap/auth service, announces services (etcd, consul, docker service links, ...), and potentially utilizes a shared filesystem (local host filesystem volume, ceph, nfs, torus, glusterfs ... )

Each component would be responsible for it's own upgrade path, installation, etc. Could be run either on a single host, ala dokku, or across a kubernetes cluster.

That's mighty different from the current sovereign architecture and I think would require a totally separate project.

ariddell:
@cromulus I agree with you in general terms. One thing to try to hang onto is the accumulated knowledge embedded in sovereign (and similar projects using ansible like debops and streisand). (Maybe @drybjed knows about new projects that fit your description?)

I'd certainly be interested in learning about projects attempting a fully modular approach. I think other users of sovereign would be interested too. If there's a mature project I think we could certainly mention it in a "See Also" section of the README.

ariddell:
Back on topic, perhaps we could add a section to the README listing mature alternatives to the roles that are currently supported in sovereign. For example, Baïkal could be mentioned as a replacement for (a part of) owncloud. Seafile and Syncthing could also be mentioned. What do you think about this @cimm?

PR #616, opened 2016-11-04
Fix bare variable depreciation

cimm:
Newer versions of Ansible complain when using bare variables like `mail_virtual_users` and prefer the newer `'{{ mail_virtual_users }}'` syntax:

depreciation

This commit fixes two of those old style bare variables: the `mail_virtual_domains` and `mail_virtual_users` one. There might be more but I did not run the full playbook.

Seen in Ansible version 2.1.2.0.


ariddell:
I've seen this too. Thanks!

Issue #611, opened 2016-10-30
How do I train rspamd?


hippich:
From my understanding sovereign moved away from dspam to rspamd. I manually removed plugin configuration from dovecot and apt-get remove dspamd itsef (had it there since debian wheezy) - Otherwise I would get errors moving messages to the Junk folder.

Now files move to spam folder just fine but I see nothing there to actually tell rspamd to look into junk folders to learn about spam.. Any pointers to how I can see if it operates normally?

SavageCore:
Use the dovecot antispam plugin to pass SPAM and HAM over to rspamc

https://kaworu.ch/blog/2014/03/25/dovecot-antispam-with-rspamd/

hippich:
So i checked - it appears that dovecot_antispam plugin is installed after sovereign is run, but plugin configuration is not. Is this is simply missed part or it was omitted on purpose?

SavageCore:
I'm not sure, I haven't actually deployed with sovereign yet. Config is easy enough and someone should make a PR!

lukecyca:
Fixed with #617. Note that detected spam still won't be automatically moved or deleted. That requires #626.

PR #617, opened 2016-11-06
Add antispam plugin config for dovecot to trigger rspamd learning whe…

hippich:
…never message is moved from inbox to spam

### DIFF
--- a/roles/mailserver/files/etc_dovecot_conf.d_90-antispam.conf
+++ b/roles/mailserver/files/etc_dovecot_conf.d_90-antispam.conf
@@ -1,0 +1,9 @@
+plugin {
+    antispam_backend = pipe
+    antispam_spam    = Junk
+    antispam_trash   = Trash
+    antispam_mail_sendmail = /usr/bin/rspamc
+    antispam_mail_spam     = learn_spam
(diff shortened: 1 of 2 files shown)


hippich:
This is related to #611

lukecyca:
This looks great. Thanks!

I am now running this patch on my personal server. I then moved a bunch of spams into the Junk folder and I hope that rspamd is using that to build it's database. Do you know how to check? I was able to bring up the Rspamd web UI, but even that couldn't tell me.

We still don't actually do anything special with identified spam, right? We need to implement [part 2](https://kaworu.ch/blog/2015/10/12/dovecot-antispam-with-rspamd-part2/) as well. Correct?

Also it seems we are using rspamd's default password of `q1` instead of [setting one](https://rspamd.com/doc/quickstart.html#setting-the-controller-password). This port isn't exposed to the Internet, so it's not an emergency. The rspamd docs suggest this is only required from "non-trusted IPs" so maybe we can just disable this?

hippich:
Yeah, i did the same and also noticed no spam appears in the junk folder. and "part 2" is indeed probably what is missing.

as for password part - not sure.. probably deserve its own ticket and PR.

PR #618, opened 2016-11-17
Fix mail_virtual_domains config reading in opendkim

nfedyashev:
without this change it fails on ansible(2.2.0.0) with the following message:

    TASK [mailserver : Generate OpenDKIM keys] *************************************
    fatal: [sovereign.host]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'unicode object' has no attribute 'name'\n\nThe error appears to have been in '/Users/nfedyashev/Projects/sovereign/roles/mailserver/tasks/opendkim.yml': line 19, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Generate OpenDKIM keys\n  ^ here\n"}

### DIFF
--- a/roles/mailserver/tasks/opendkim.yml
+++ b/roles/mailserver/tasks/opendkim.yml
@@ -14,11 +14,11 @@
 
 - name: Create OpenDKIM key directories
   file: state=directory path=/etc/opendkim/keys/{{ item.name }} group=opendkim owner=opendkim
-  with_items: mail_virtual_domains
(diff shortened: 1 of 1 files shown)


nfedyashev:
I'm not an expert in ansible but that update allowed me to continue running this playbook

jplock:
There are a few other places in the roles where we need to fix this too. Thanks!

PR #619, opened 2016-11-20
Bump selfoss_version

nfedyashev:
[Changelog](https://github.com/SSilence/selfoss/releases)

I'm running php7 so without this update selfoss 2.15 is basically unusable.
As a bonus it has other features and bug fixes.


jplock:
Thanks!

PR #620, opened 2016-11-20
Enable ownCloud virtual host automatically

nfedyashev:
Just noticed that owncloud playbook does not enables itself as apache virtual host. This is a bit surprising because roundcube and selfoss activates themselves just fine. This PR fixes it.

`roles/webmail/tasks/roundcube.yml` was used as inspiration for this PR.


neuhaus:
Thanks!

PR #634, opened 2016-12-10
Clarifies environments, adds non-passwordless option

OKNoah:
It was unclear where do to the commands. Not everyone will know ansible should not be run from the server. Also, the "optional" passwordless option was not actually optional, given how this was written.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,9 +106,9 @@ Your new account will be automatically set up for passwordless `sudo`. Or you ca
 
 ## On your local machine
 
-Ansible (the toll setting up your server) runs locally on your computer and sends commands to the remote server. Download this repository somewhere on your machine, either through `Clone or Download > Download ZIP` above, `wget`, or `git` as below
+Ansible (the tool setting up your server) runs locally on your computer and sends commands to the remote server. Download this repository somewhere on your machine, either through `Clone or Download > Download ZIP` above, `wget`, or `git` as below
     
-    git https://github.com/sovereign/sovereign.git
+    git clone https://github.com/sovereign/sovereign.git
 
 ### 4. Configure your installation
 
--- a/README.md
+++ b/README.md
@@ -63,6 +63,10 @@ You do not need to acquire an SSL certificate.  The SSL certificates you need wi
 Installation
 ------------
 
+## On the remote server
+
+The following steps are done on the remote server by `ssh`ing into it and running these commands.
+
 ### 1. Install required packages
 
     apt-get install sudo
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,11 +100,19 @@ Authorize your ssh key if you want passwordless ssh login (optional):
...
     echo 'deploy ALL=(ALL) NOPASSWD: ALL' > /etc/sudoers.d/deploy
 
-Your new account will be automatically set up for passwordless `sudo`.
+Your new account will be automatically set up for passwordless `sudo`. Or you can just add your `deploy` user to the sudo group.
+
+    adduser deploy sudo
+
+## On your local machine
+
+Ansible (the toll setting up your server) runs locally on your computer and sends commands to the remote server. Download this repository somewhere on your machine, either through `Clone or Download > Download ZIP` above, `wget`, or `git` as below
```
lukecyca: I'm not familier with this use of the word "toll". What does it mean in this context?

README.md
```diff
@@ -96,11 +100,19 @@ Authorize your ssh key if you want passwordless ssh login (optional):
...
+    
+    git https://github.com/sovereign/sovereign.git
```
lukecyca: I believe we need the `clone` command in here

OKNoah:
@lukecyca Fixed those typos.

ariddell:
LGTM. I'd still like to drop entirely the explicit step where one must create a `deploy` user. It would make things simpler. If there's ever consensus on this, perhaps we could switch over.

README.md
```diff
@@ -172,7 +184,9 @@ First, make sure you’ve [got Ansible 1.9.3+ installed](http://docs.ansible.com
 
 To run the whole dang thing:
 
-    ansible-playbook -i ./hosts site.yml
+    ansible-playbook -i --ask-sudo-pass ./hosts site.yml
+    
+If you chose to make a passwordless sudo deploy users, you can omit the `--ask-sudo-pass` argument.
```
sigmafeud: deploy users -> deploy user

ariddell:
Apparently I'm not the only one who thinks having the deploy user isn't necessary:

https://github.com/google/fury/blob/master/README.opinions.md#just-log-in-as-root-already

```
Just log in as root already

Excluding the case of using automation for specific types of service deployment, your configuration needs root@machine anyway. Stop pretending you don't, and don't litter your config mgmt system with the twelve different mechanisms for privilege escalation that people think they need. Just bite the bullet and let the config mgmt log in as root.

Admittedly, this does exclude people who have valid reasons for restricting privilege escalation. But they're already dealing with headaches stemming from that anyway, so I'm just going to give them one more.
```

lukecyca:
@ariddell sounds good. Open a new issue or pull request for that!

PR #648, opened 2017-02-01
Various fixes and tweaks

alexec:
This should not be merged whole, only piecemeal (I've commented out `autoconfig` and various other roles). This works on a Digital Ocean Ubuntu server.

* Added feature flags to customise your set-up.
* Fixed `fail2ban` set-up
* Install python (as Ansible needs this)
* Added some mail troubleshooting
* Fixed OpenDKIM so it starts
* Fixed SelfOSS auto-update

Theres some problems with OpenDMARC cronjob (basically newer versions of MySQL are stricter).


bsima:
> Removed mosh (a bit of an unknown and could be another hole for hackers to poke around)
 
:( I love mosh! FWIW mosh just uses ssh, so it's presumably as secure as ssh. See the two security questions in the FAQ (ctrl+f "security"): https://mosh.org/

jplock:
@alexec is it possible to submit separate PR's for some of these items to make it easier to review and merge in pieces?  Some of this looks good.

alexec:
kk, I'll break this up and submit separately

jplock:
Thanks!

PR #652, opened 2017-02-11
Updated the name of the fail2ban jails

alexec:


jplock:
Why would we want to do this?

alexec:
On Ubuntu 16.04, fail2ban would not start due to the jail names. They did not match the actual names of the jails in `/etc/fail2ban/filter.d`.

lukecyca:
I tested this PR on Debian 8.7 and it breaks, so I looked into it a bit...

The "section name" in the config file is not necessarily the filter name. If you look in `/etc/fail2ban/jail.conf`, there is a section called `ssh` which is configured to use the `sshd` filter, which it finds in `/etc/fail2ban/filter.d`.

Not sure what problem you hit with Ubuntu, @alexec. Perhaps you can let us know what your `/etc/fail2ban/jail.conf` looks like.

mikeashley:
I am closing this PR since it appears to have gone stale after failing review.

PR #653, opened 2017-02-11
Update the defaults file for OpenDKIM to include the SOCKET so that t…

alexec:
…he service starts


mikeashley:
What problem does this solve?  Why is this patch needed?

mikeashley:
Obsoleted by PR #720

PR #654, opened 2017-02-11
Updated scripts to use raw mysql command as the Ansible mysql module …

alexec:
…does not work with Ubuntu 16


mikeashley:
No longer needed because of #720.

PR #655, opened 2017-02-11
Use php* packages rather than php5* for packages on Ubuntu

alexec:
Also invokes the update PHP file, rather that the HTTP endpoint (which is secured and therefore cannot be invoked this way).

### DIFF
--- a/roles/news/tasks/selfoss.yml
+++ b/roles/news/tasks/selfoss.yml
@@ -21,9 +21,9 @@
 - name: Install selfoss dependencies
   apt: pkg={{ item }} state=present
   with_items:
-    - php5
-    - php5-pgsql
-    - php5-gd
+    - "{{ (ansible_distribution == 'Ubuntu') | ternary('php', 'php5') }}"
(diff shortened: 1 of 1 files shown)


jplock:
Will this work in both ubuntu and debian?

alexec:
Yes. This will work for both.

lukecyca:
This looks fine to me but I don't use Selfoss so can't really test. Can anyone else?

Also worth noting that PHP packages are installed from a variety of places. Not just here.

 - ./roles/mailserver/tasks/z-push.yml
 - ./roles/readlater/tasks/wallabag.yml
 - ./roles/webmail/tasks/roundcube.yml

alexec:
I know it's a bit of change of direction for this project, but I'd  could be much easier to run and main docker builds.

lukecyca:
> I know it's a bit of change of direction for this project, but I'd could be much easier to run and main docker builds.

Certainly. If you're willing to lead that, I could create a new repo for you.

alexec:
Perhaps to start with, we could discus this on a new issue?

PR #656, opened 2017-02-11
Use wget to add apt-key for OwnCloud rather that apt_key module.

alexec:


alexec:
The URL parameter fails SSL verification checks, `wget` does not. This is a problem on both Ubuntu and Debian.

jplock:
I haven't experienced the SSL verification issue on Trusty.

mikeashley:
Likewise, I have not had any problem with SSL verification checks.  Is this something you can reproduce in a test environment?

mikeashley:
Closing this since it does not appear to be a needed change.  Please resubmit if this is reproducible in a test environment.

PR #657, opened 2017-02-11
Minor fixes to OpenVPN set-up

alexec:


roles/vpn/templates/etc_openvpn_server.conf.j2
```diff
@@ -187,7 +187,7 @@ ifconfig-pool-persist ipp.txt
 # (The OpenVPN server machine may need to NAT
 # or bridge the TUN/TAP interface to the internet
 # in order for this to work properly).
-push "redirect-gateway def1"
+push "redirect-gateway def1 bypass-dhcp"
```
jplock: What does this do?

alexec: From https://www.digitalocean.com/community/tutorials/how-to-set-up-an-openvpn-server-on-ubuntu-16-04:

> The settings above will create the VPN connection between the two machines, but will not force any connections to use the tunnel. If you wish to use the VPN to route all of your traffic, you will likely want to push the DNS settings to the client computers.

alexec: From https://www.digitalocean.com/community/tutorials/how-to-set-up-an-openvpn-server-on-ubuntu-16-04:

> The settings above will create the VPN connection between the two machines, but will not force any connections to use the tunnel. If you wish to use the VPN to route all of your traffic, you will likely want to push the DNS settings to the client computers.

jplock: I think we want all traffic to traverse the VPN

jplock: If you can revert this line, I think the rest is ok to merge.

PR #660, opened 2017-02-24
Avoid opportunity for copy-paste trouble

JasonGhent:
Also more scriptable, in vein of alternative commands provided


ariddell:
Maybe add a note about what the ``sed`` command does for those who might not be familiar with it?

JasonGhent:
Sure. sed is used here to truncate the hash type from the beginning of the doveadm output.

ariddell:
Right, perhaps add such a note to the README.md? It's a good change. Thanks for this.

JasonGhent:
Updated. 😄

JasonGhent:
Any additional feedback?

ariddell:
Looks great! Thanks.

PR #668, opened 2017-03-21
Fix cgit

Titan-C:
I was testing the git hosting role, and although the cgit website displayed neither the about page of a repository displayed any content nor was it possible to view the content of any file.

- The first problem originated because my readme file is README.md a markdown file. That dependency is not installed with pip.
- Second problem is that source files are parsed by the syntax highlighter. That script in cgit enforces the use of python3 but the pygments dependency is installed with the system pip which in debian jessy is still python2, thus pygments was not available for use in the python3 interpreter. I have added a Python3 pip and install pip3 dependencies for it

I was also thinking it might be good to update cgit version to current 1.1. That version even enforces python3 in the markdown converter. That would allow to remove the pip2 install dependency that I for the moment have left in the role. Also I don't know how to force the install step once the version has been updated.

### DIFF
--- a/roles/git/tasks/cgit.yml
+++ b/roles/git/tasks/cgit.yml
@@ -4,17 +4,9 @@
   with_items:
     - groff
     - libssl-dev
-    - python-pip
     - python3-pip
   tags:
     - dependencies
-
-- name: Install cgit pip dependencies
-  pip: name={{ item }}
-  with_items:
-    - docutils
-    - pygments
-    - markdown
 
 - name: Install cgit pip dependencies python 3
   pip:
--- a/roles/git/tasks/cgit.yml
+++ b/roles/git/tasks/cgit.yml
@@ -5,6 +5,7 @@
     - groff
     - libssl-dev
     - python-pip
+    - python3-pip
   tags:
     - dependencies
 
@@ -13,6 +14,16 @@
   with_items:
     - docutils
     - pygments
+    - markdown
+
+- name: Install cgit pip dependencies python 3
+  pip:
+    name: "{{ item }}"
+    executable: pip3
+  with_items:
+    - docutils
+    - pygments
+    - markdown
 
 - name: Download cgit release
   get_url: url=http://git.zx2c4.com/cgit/snapshot/cgit-{{ cgit_version }}.tar.xz
--- a/roles/git/defaults/main.yml
+++ b/roles/git/defaults/main.yml
@@ -1,3 +1,3 @@
-cgit_version: 0.12
+cgit_version: 1.1
 cgit_domain: "git.{{ domain }}"
 gitolite_version: 3.6.4
--- a/roles/git/tasks/main.yml
+++ b/roles/git/tasks/main.yml
@@ -1,2 +1,4 @@
-- include: gitolite.yml tags=gitolite
-- include: cgit.yml tags=cgit
+- include: gitolite.yml
+  tags: gitolite
+- include: cgit.yml
+  tags: cgit


mikeashley:
Looks like cgit switched from Python 2 to Python 3 between versions 0.12 and 1.1.

In general, we don't try to support upgrades in the Ansible scripts.

I like the idea of upgrading to version 1.1.  Could you update this pull request to simply install 1.1 and install the correct dependencies for that version of cgit?

roles/git/tasks/cgit.yml
```diff
@@ -5,6 +5,7 @@
     - groff
     - libssl-dev
     - python-pip
```
mikeashley: Is the `python-pip` package still needed?

Titan-C: no, just forgot about it.

Titan-C:
How can I force the update to version 1.1?
The sources are downloaded but the "build and install cgit" is not executed because the target file of version 0.12 is already present.

mikeashley:
You'll need to patch it up by hand, e.g., delete the directory `/var/www/htdocs/cgit`.  Looks like some subsequent tasks will need similar treatment.

We install two files from templates: `/etc/cgitrc` and `/etc/apache2/sites-available/cgit`.  Have you reviewed those to see if they need to be changed to match what is included with or recommended by cgit 1.1?

Titan-C:
> We install two files from templates: /etc/cgitrc and /etc/apache2/sites-available/cgit. Have you reviewed those to see if they need to be changed to match what is included with or recommended by cgit 1.1?

I'm very ignorant on web services. From what I can figure out `/etc/cgitrc` does not require any change `/etc/apace2/sites-availabe/cgit` has a different configuration in the cgit readme, relevant would be this changes 
```diff
@@ -12,7 +12,8 @@
     <Directory "/var/www/htdocs/cgit/">
         AllowOverride None
         Options +ExecCGI
-        Require all granted
+        Order allow,deny
+        Allow from all
     </Directory>
 
     Alias /cgit.png         /var/www/htdocs/cgit/cgit.png
```
Shall I follow the cgit suggeted config?

mikeashley:
I agree on the changes to the Apache configuration.

I did not author the original cgit install in Sovereign, so I am not sure what is different.  What I do is diff the old one with the new one and think about what changes make sense.  To help reviewers today and maintainers in the future, I have started first committing a pristine copy of the configuration file and then in a second commit making the sovereign-specific changes.

If it were me, I would follow the same pattern here.  Take a clean copy of cgit 1.1's cgitrc, commit it, and then commit the modifications.

Titan-C:
> I agree on the changes to the Apache configuration.

Sorry I just got very confused. You mean you agree on the configuration that sovereign currently has: that means keep the red line from my previous diff. Or you agree with cgit current default from their readme: that means the green lines of my previous diff.

Titan-C:
> If it were me, I would follow the same pattern here. Take a clean copy of cgit 1.1's cgitrc, commit it, and then commit the modifications.

Having a more careful look at git blame on the cgit codebase. Neither their readme nor their cgitrc.5.txt have substantially changed in the last 2 years and cgit v0.12 only 15 months old.  Thus my best guess is that sovereign's current configuration of this files is good and does not need changes.

mikeashley:
Thanks for asking, and I double-checked.  Actually, we should keep the red lines (what sovereign already has).  In Apache 2.4, the require directive [is favored](https://httpd.apache.org/docs/2.4/upgrading.html) over the old order/allow/deny directives.

Titan-C:
Then is all set? Should I write somewhere that cgit got updated to a new version and that the user has to manually delete `/var/www/htdocs/cgit` for the new version to be installed?

mikeashley:
Did you review cgitrc?

Titan-C:
> Did you review cgitrc?

Yes and I find sovereign's current config more reasonable that cgit example.

mikeashley:
I went ahead and merged.  Thank you!  I am not sure of a good place to make that note for users upgrading.  Personally, I take a quick look at the changes to roles I am updating before I rerun the playbook and then check it afterwards.  That's enough for me.

If it's not enough, perhaps we could make notes on the wiki?  I don't think cluttering the README or commit logs is the right place.

PR #670, opened 2017-04-06
Implementing password hashing inside password_hash filter plugin

synchrone:


ariddell:
This is interesting. This does mean that the passwords would be in the clear on disk, right? (Not a criticism, just a remark.)

synchrone:
Only on control host, which already knows all the other passwords. 
The configured result for znc and mailserver is the same as before, a hash in the config file and the DB respectively
-- 
Best Regards, 
Aleksandr Bogdanov

synchrone:
Any feedback here? Would be really nice to reduce manual work for the user

ariddell:
I like this. How about copying-and-pasting the text removed from the README into the docstrings of the filters so everyone knows what is happening? That way it's crystal clear what's going on.

neuhaus:
I love this! I had been thinking about how we need to simplify the setup process, especially the hassle with all those password hashes.

neuhaus:
Great contribution, keep it up! This simplifies setup a lot.

PR #672, opened 2017-04-07
fix pgsql connection issues from missing quotes in selfoss config file

srccon:
I found that some of the generated password characters can cause issues with selfoss connecting to the pgsql database, adding quotes in the config fixes this issue.

### DIFF
--- a/roles/news/templates/var_www_selfoss_config.ini.j2
+++ b/roles/news/templates/var_www_selfoss_config.ini.j2
@@ -1,9 +1,10 @@
 ; see https://github.com/SSilence/selfoss/wiki/Configuration for more information about the configuration parameters
 [globals]
 db_type=pgsql
-db_database={{selfoss_db_database}}
-db_username={{selfoss_db_username}}
-db_password={{selfoss_db_password}}
+db_host=127.0.0.1
+db_database="{{selfoss_db_database}}"
+db_username="{{selfoss_db_username}}"
+db_password="{{selfoss_db_password}}"
 db_port=5432
 db_prefix=
 logger_level=DEBUG


roles/news/templates/var_www_selfoss_config.ini.j2
```diff
@@ -1,9 +1,10 @@
 ; see https://github.com/SSilence/selfoss/wiki/Configuration for more information about the configuration parameters
 [globals]
 db_type=pgsql
-db_database={{selfoss_db_database}}
-db_username={{selfoss_db_username}}
-db_password={{selfoss_db_password}}
+db_host=127.0.0.1
```
mikeashley: Why this change?

roles/news/templates/var_www_selfoss_config.ini.j2
```diff
@@ -1,9 +1,10 @@
...
+db_database="{{selfoss_db_database}}"
```
mikeashley: Do you think quoting the database and username are necessary?  Just trying to focus the patch on the specific problem and not make speculative changes.

roles/news/templates/var_www_selfoss_config.ini.j2
srccon: Technically it's not necessary, I posted up what I had changed to get it to work. I like to over specify.

roles/news/templates/var_www_selfoss_config.ini.j2
srccon: I'm not sure what the escape character are, commas certainly. Prevents against escape characters causing issues in the username as well, in case a user randomly generates their username.

srccon: Database, probably not.

roles/news/templates/var_www_selfoss_config.ini.j2
srccon: It defaults to localhost anyway, this would make it easier for someone to specify a separate db server.

roles/news/templates/var_www_selfoss_config.ini.j2
fengor: quoting those is considered good form. I'd +1 those changes

roles/news/templates/var_www_selfoss_config.ini.j2
neuhaus: The default is "localhost", not 127.0.0.1. This change could mean that the database connection uses a TCP socket instead of a UNIX domain socket which is twice as fast. Please either change it to localhost or remove it.

mikeashley:
I removed the `db_host` setting and merged the (squash) commit.  Thanks for the contribution!

PR #677, opened 2017-04-09
Use postmaster@ for DMARC reports

### LINKED ISSUE #676: opendmarc sends reports with personal email
I get bounces every once in a while from DMARC reports. That's normal, but it's a little concerning because they are to my personal email rather than an admin email like `postmaster@domain.name`. This is because we're setting `--report-email` as the user's address when it should probably postmaster@.

I would consider this an information leak because a spammer could send spam to any address on your domain and the DMARC report would include your real email address in the report back to the spammer's mail server.

adborden:
Generate the [DMARC report](https://www.mankier.com/8/opendmarc-reports#--report-email) from postmaster@ instead of personal email address.

Fixes #676


neuhaus:
Thanks!

PR #679, opened 2017-04-09
Restart prosody after LE cert renewal

### LINKED ISSUE #678: Prosody has expired LE cert after initial expiration
Prosody needs to be restarted to pick up the new LE certs after renewal.

adborden:
Fixes #678


neuhaus:
thanks!

Issue #538, opened 2016-03-17
Upgrade to Ansible 2.0


ariddell:
The Ansible 2.0 refactor was significant. It would be useful to have a version of sovereign that's compatible with the current version of ansible.
- [ ] Get Ansible 2.0 running in the jessie branch
- [ ] Revert workarounds for pre-ansible 2.0 (#534 ?)

neuhaus:
This issue is getting more important, Xenial ships with Ansible 2.0

PR #684, opened 2017-04-13
Switching to Ansible 2

synchrone:
So it's been a year since https://github.com/sovereign/sovereign/issues/538.
Sovereign runs well, except for one `shell:` invocation acting up with ansible 2.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-ansible>=1.9.3,<2
+ansible>=2.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-ansible>=2.2
+ansible==2.2.2.0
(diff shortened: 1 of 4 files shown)


ariddell:
Thanks for this. I'd be in favor of merging this. What little I do with Ansible is with 2.0

synchrone:
okay, so who else needs to review this?
I'd like to have this merged, since it would enable a prospective route53/google cloud dns role, as those modules are only available in Ansbile 2.2+

ariddell:
Anyone else have thoughts? This is a big change.

ariddell:
OK. In the spirit of http://rfc.zeromq.org/spec:42/C4 I'm going to merge this. If anyone's upset, submit a PR reverting it.

Issue #689, opened 2017-04-29
Modify permissions to allow ssl-cert group access


themsay:
TASK [common : Modify permissions to allow ssl-cert group access] **********************************************************************************
fatal: [158.81.196.11]: FAILED! => {"changed": false, "failed": true, "gid": 0, "group": "root", "mode": "0700", "msg": "chgrp failed: failed to look up group ssl-cert", "owner": "root", "path": "/etc/letsencrypt/archive", "size": 4096, "state": "directory", "uid": 0}
	to retry, use: --limit @/Users/ndu/Projects/Mail/sovereign/site.retry

PR #690, opened 2017-04-29
Add group name ssl-cert for SSL certificates

themsay:
Based on issue https://github.com/sovereign/sovereign/issues/689


neuhaus:
thanks!

Issue #704, opened 2017-08-05
rspamd repository unavailable on arm machines


neuhaus:
When trying to use sovereign on a raspberry pi 3 I ran into the problem that the rspamd repository doesn't offer binaries for the "armhf" architecture.

On a Raspberry Pi 3 with Raspbian 8, I can see the following Ansible Variables that we can use to detect this platform:

        "ansible_machine": "armv7l",
        "ansible_architecture": "armv7l",

There is the stock rspamd package provided by Raspbian (jessie) but it's still version 0.6.10.

@carljm  Whats the minimum version of rspamd required by our setup?

carljm:
@neuhaus I don't have any idea, sorry :/

PR #705, opened 2017-08-05
Add repo for rspamd on ARMv7. 

neuhaus:
Use yunohost repository for more recent rspamd package (currently at 1.3.5) on
ARMv7 (armhf, e.g. Raspberry Pi 3). Fixes issue #704

### DIFF
--- a/roles/mailserver/tasks/rspamd.yml
+++ b/roles/mailserver/tasks/rspamd.yml
@@ -3,11 +3,25 @@
 
 - name: Ensure repository key for Rspamd is in place
   apt_key: url=https://rspamd.com/apt-stable/gpg.key state=present
+  when: ansible_architecture != "armv7l"
+  tags:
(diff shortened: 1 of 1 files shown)


mikeashley:
What's the reasoning behind moving this upstream?  It seems like if we accept this change we are tacitly supporting ARM.  Are there other changes needed to support ARM?  If so, should we take them all?

neuhaus:
I installed Raspbian on a Raspberry Pi 3 and tried it with Sovereign and this was the part that broke. The rest looked ok (tested only email).
I wouldn't say we officially support ARM, but I see not reason not to fix the breakage when I encounter it.

neuhaus:
By the way, why does the README state that a 64bit version of Linux is required? Does Sovereign not work on 32bit? If so, why not? (it appears to work on 32bit on ARM).

Issue #685, opened 2017-04-17
Roundcube sieve filters not being applied


bmiller59:
I have added some new filters to the managesieve filter group. I removed the "Stop evaluating rules" line in the default spam rule and added my rules after that. None of the filters are being applied, either in roundcube or in my remote imap client. All of the filters are enabled (not disabled). Is there some other configuration that is neeed to get them working? Or does some service need to be restarted, etc? Thank you.

bmiller59:
I was able to fix this problem by making the following change to the file /etc/dovecot/conf.d/20-lmtp.conf

Replace this existing line: #mail_plugins = $mail_plugins
With this: mail_plugins = $mail_plugins sieve

And then restart dovecot with: sudo service dovecot restart

It seems LMTP is being used, but sieve was not enabled for it. This change enables it, and then sieve processing works as expected.

Issue #691, opened 2017-05-01
etc_dovecot_conf.d_90-plugin.conf is orphaned


Xeboc:
As of this commit:

https://github.com/sovereign/sovereign/commit/ecaa4c23308adbdb5d76937296090ca7a6cc5944#diff-60a2d80cf2f6bad5c7a33594b338f4aa

When dspam.yml was removed, 90-plugin.conf is orphaned and never copied over.  Is it needed for solr functionality?  I'm just diving into this project, so I haven't tested the solr functionality myself, and was hoping a maintainer knew.

https://github.com/sovereign/sovereign/blob/master/roles/mailserver/files/etc_dovecot_conf.d_90-plugin.conf

neuhaus:
paging @al3x

mikeashley:
Fixed with PR #716.

PR #708, opened 2017-08-18
Copy missing configuration file "20-imap.conf"

Terr:
I noticed that my Sieve filters were not being applied. The [issue](https://github.com/sovereign/sovereign/issues/685) reported by @bmiller59 pointed me to a `lmtp` section that was missing from the Dovecot configuration files.

Apparently the file containing the missing setting was already present in the repository, it was just not being transfered by the dovecot Ansible task. So I added `20-imap.conf` to the list of files to copy.

### DIFF
--- a/roles/mailserver/tasks/dovecot.yml
+++ b/roles/mailserver/tasks/dovecot.yml
@@ -59,6 +59,10 @@
   template: src=etc_dovecot_conf.d_15-lda.conf.j2 dest=/etc/dovecot/conf.d/15-lda.conf
   notify: restart dovecot
 
+- name: Template 20-imap.conf
+  template: src=etc_dovecot_conf.d_20-imap.conf.j2 dest=/etc/dovecot/conf.d/20-imap.conf
+  notify: restart dovecot
+
 - name: Template dovecot-sql.conf.ext
   template: src=etc_dovecot_dovecot-sql.conf.ext.j2 dest=/etc/dovecot/dovecot-sql.conf.ext
   notify: restart dovecot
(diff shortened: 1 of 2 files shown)


mikeashley:
It looks like this bug has been here always.  Furthermore, the copy of `20-imap.conf` on my server is identical to the copy held by Sovereign.  This suggests Sovereign's copy is the distribution's copy.  Should we instead delete `20-imap.conf` from our configuration?

mikeashley:
Issue #691 observes an identical problem with `90-plugin.conf` being orphaned when `dspam.conf` was dropped with commit ecaa4.  Commit log indicates this was part of swapping in Rspamd for dpsam.

neuhaus:
looks good to me, have you tested this @Terr ?

Terr:
@neuhaus I did, only to discover that there was a `{{ domain }}` variable in the file. I changed my commit to process that config as a template.

I tested it out with the testing Vagrant setup and the file is copied and filled in correctly

neuhaus:
Good. What happens with existing installations?

Terr:
@neuhaus Good question. The sieve filters should start to work again, but I don't know what the other `mail_plugins` in the `imap` protocol section will do.

Those other plugins (`antispam`, `fts`, `fts_solr`) are listed in other Dovecot config files but have not been loaded for as long as `20-imap.conf` has been missing. I don't know what the consequences are of having them enabled suddenly.

mikeashley:
I've incorporated this patch into #716 and applied it on my production server.  It seems to work, but I am not a heavy user of sieve filters.

PR #714, opened 2017-10-26
Update for Debian 9

bryanvoss:


ariddell:
I'm assuming this isn't a problem for Jessie (Debian 8) since python3-software-properties exists.

Issue #661, opened 2017-02-28
Rspamd error when moving spam messages


iboxifoo:
Ever since the switch to using Rspamd (possibly following #626), whenever I attempt to move messages either into the junk folder or wrongly flagged spam out of the junk folder I see this message:

> The IMAP command “UID COPY” (to Junk) failed for the mailbox “INBOX” with server error: antispam signature not found.

I'm using Apple Mail on latest Sierra. I've poked around the logs but not sure where to find anything on the server that would help debug. Some searches online suggest a permission issue with the Rspamd binary but I'm not sure where to begin.

If anyone can give me pointers I'd be grateful. Or maybe this is a problem seen by others. Thanks.

lukecyca:
FWIW I am also using Apple Mail on Sierra, with no issues when moving messages in/out of the junk folder.

I do get an error like that if, for example, I try to move a message from a different one of my accounts (say gmail) into the junk folder of my sovereign account.

To troubleshoot, you could connect to rspamd's web console and ensure that it is analyzing and creating signatures for the messages in question.

mikeashley:
I'm seeing this also, but I have not had time to find the root cause.

mikeashley:
Pull request #716  fixes this.

Issue #647, opened 2017-01-31
Sieve not configured properly


JGjorgji:
15-lda.conf says the settings apply to lmtp too however mail is not processed by sieve rules, this needs to be added so it works:

```
protocol lmtp {
  # Space separated list of plugins to load (default is global mail_plugins).
  mail_plugins = $mail_plugins sieve
}
```

I've added this in a separate lmtp conf file and it works fine, also the firewall needs to open a port for managesieve, port 4190 which is not opened.

neuhaus:
If you manage sieve using the sieve plugin for Roundcube you don't need to open the port in the firewall so it makes sense to keep it that way as a secure default.

JGjorgji:
That makes sense guess i was using thunderbird and didn't consider that. It would be useful to have this mentioned someplace in case someone wants to use it in this way too.

How about the filtering is it working for you by default?

neuffer:
On 31.01.2017 20:14, Gjorgji Jankovski wrote:
>
> That makes sense guess i was using thunderbird and didn't consider
> that. It would be useful to have this mentioned someplace in case
> someone wants to use it in this way too.
>
>

Maybe this information could be added to the project wiki ?

bjdeliduka:
I just did a fresh install of sovereign on a linode running Debian 8 and noticed the following:

 - roundcube lacks a method of changing the mailbox/user password.

 - there is a filter set named "managesieve" with a filter named #1 that checks message headers for "x-dspam-result" values matching a regular expression.  matches are to be marked as **Read** and moved to a folder named **Spam**
 - the **Spam** folder does not exist
 - messages do not contain "x-dspam-result" 
 - test email messages from http://www.emailsecuritycheck.net/ all pass through to the inbox without any flagging as spam not warnings about "viruses" that should be triggered by the test content.

I have not checked the above change to see if it fixes the spam / sieve issue.  (manually creating the Spam folder does not have any impact on the functionality of this. Saving the filter requires setting a valid name (it is blank by default)

Did I miss something?

neuhaus:
The only supported way to change a user's mail password is to change it in the `group_vars/sovereign` file. If you change it in some other way it will get overwritten the next time sovereign is run.

Given that the spam folder doesn't exist - will sieve create it when moving a message into it?

dspam was replaced by rspam. This configuration needs to be updated.

Issue #637, opened 2016-12-20
Use rspamd/rmilter for greylisting; remove Postgrey


lukecyca:
[Rspamd's greylisting module](https://rspamd.com/doc/modules/greylisting.html) is enabled by default in our configuration, so it is already performing greylisting. We previously used (and still use) Postgrey as well. I think this means we are currently greylisting _twice_.

We need to verify that Rspamd's greylisting is working as intended, and then remove Postgrey completely.

Dosenpfand:
I don't think that the rspamd greylisting module is active by default on Debian 8 (and therefore at the moment when using sovereign).
What i did to check:

- Disable postgrey by removing
`check_policy_service inet:127.0.0.1:10023,` from `roles/mailserver/templates/etc_postfix_main.cf.j2 => All mails arrived instantly, so apparantly no greylisting active
- Search for the textstring `greylist` `in /etc/rspamd/*` => No appropriate results.

lukecyca:
@Dosenpfand thanks for investigating. I would really like to see us move away from postgrey and then ensure that rspamd's greylisting is enabled and working.

If you want to do more work towards this or take a stab at a PR, I'll happily help test it on my own installation and help get it merged.

Dosenpfand:
@lukecyca Unfortunately I will not have time to work in this in the foreseeable future. Additionally I kept disabled postgrey the last month and did not see any spam hitting my inbox.

mikeashley:
With rspamd 1.6 the greylisting module is active by default.  See `/etc/rspamd/modules.d/greylist.conf`.  I am turning off postgrey and will report back what I find in a couple of weeks.

mikeashley:
Fixed with PR #716.

PR #716, opened 2017-11-05
Correct mailserver configuration

mikeashley:
Some configuration files were not getting installed.  Also, rmilter is no longer needed with the rspamd 1.6.


mikeashley:
This pull request reinstates solr for fast searching, but it looks like it is broken.  GET requests are returning 500s (internal server error).  I'm investigating.  Recommend holding off merging until I get to the bottom of it.

mikeashley:
The solr error was a goose chase.  Solr is working fine with dovecot, verified with Solr's web interface and both dovecot's and tomcat's logs when searching using Roundcube.

mikeashley:
I updated this PR after merging #708 .  It's baked for two weeks now on my production server with no problems, and Iv'e been reviewing rspamd's logs using its admin web UI.  I think this is ready to merge.

mikeashley:
This PR addresses issues #637, #661, and #691.

neuhaus:
Looks good. One thing i noticed is that the DoveCot AntiSpam plugin has been deprecated.
See https://wiki2.dovecot.org/Plugins/Antispam
Is this something we should be doing instead?

mikeashley:
I noticed that too.  We've got a couple of issues with sieve (#647, #608).  I was thinking of tackling that separately once I get through this PR and the two behind it to address DMARC and DKIM setup.

Issue #715, opened 2017-11-05
Installation fails due to missing owncloud folder


der-gee:
Hello everyone,

i'm trying to get a new installation running. Everything works like a charm, expect the owncloud installation:

```
TASK: [owncloud | Move ownCloud data to encrypted filesystem] ***************** 
failed: [my-host.tld] => {"changed": true, "cmd": ["mv", "/var/www/owncloud/data", "/decrypted/owncloud-data"], "delta": "0:00:00.014927", "end": "2017-11-05 04:03:43.970496", "rc": 1, "start": "2017-11-05 04:03:43.955569", "warnings": []}
stderr: mv: cannot stat ‘/var/www/owncloud/data’: No such file or directory
FATAL: all hosts have already failed -- aborting
```

If I create said folder by myself and start the playbook again, the installation completes, but I'll get an kind of folder like view from apache visiting the site https://cloud.my-host.tld without any content.

When the installation is done, the folder:
`/var/www/owncloud/data` contains a link to `/decrypted/owncloud-data`, which is empty 

`dpkg` lists owncloud as installed.

Can anybody provide some suggestions what could be wrong?

Thanks in advance,
der-gee

der-gee:
I messed up and have to wait a week until further certificates can be issued :)

I think I have to change the package from owncloud to owncloud-files in the owncloud.yml.

I'll close this issue (and reopen it, if I was wrong).

der-gee:
Hello again,

sorry, but I'm to tired to figure out how to create a pull-request.

But I've managed to complete a installation in one run with changeing owncloud to owncloud-file.

Here's the link which led me to this conclusion: [owncloud manual](https://doc.owncloud.org/server/9.0/admin_manual/installation/linux_installation.html)

```/master/roles/owncloud/tasks/owncloud.yml 
- name: Install ownCloud
  apt: pkg=owncloud-files update_cache=yes```

Worked for me.

der-gee:
If this was a finding, it should be solved with PR #718

neuhaus:
Let's leave the issue open until the PR has been merged.
What Linux distribution (and version) are you using?

der-gee:
Distributor ID: Debian
Description:    Debian GNU/Linux 8.9 (jessie)
Release:        8.9
Codename:       jessie

[...]

PR #718, opened 2017-11-12
Changed owncloud package to owncloud-files. Installation without -files failed

der-gee:
see issue #715 

Please review and let me know what you think.

### DIFF
--- a/roles/owncloud/tasks/owncloud.yml
+++ b/roles/owncloud/tasks/owncloud.yml
@@ -28,7 +28,7 @@
     - dependencies
 
 - name: Install ownCloud
-  apt: pkg=owncloud update_cache=yes
+  apt: pkg=owncloud-files update_cache=yes
   tags:
     - dependencies
 


mikeashley:
If you install the owncloud metapackage, how does it fail?  Reading the manual, I think it should have worked.  Was there a dependency conflict of some kind with Apache or some other prereq?

der-gee:
I really can't tell. I tried it severall times and even the -vvvv command let me haning there clueless. I've stated the only viable error-log in the linked issue as the behaviour leading me to my conclusion:

I've started reading the [documentation](https://doc.owncloud.org/server/latest/admin_manual/installation/linux_installation.html) and at some point found other people in comparable situations who switched to *-files, reporting that it solved their issues.

Since sovereign takes care of the prereqs/dependencies needed for owncloud, I thought I give it a try and ommit these by stating the -files option, which worked perfect for me.

PR #727, opened 2017-12-09
Upgrade Selfoss to version 2.17

mikeashley:
This picks up a variety of improvements but most importantly makes
selfoss [compatible with Apache
2.4](https://github.com/SSilence/selfoss/commit/8bc38c0820afb8697bb729d4dfd7165db376084a).

Selfoss uses composer to install dependencies starting with version
2.17.  Task list updated accordingly.

### DIFF
--- a/roles/news/defaults/main.yml
+++ b/roles/news/defaults/main.yml
@@ -6,7 +6,7 @@ selfoss_domain: "news.{{ domain }}"
 selfoss_db_username: selfoss
 selfoss_db_password: "{{ lookup('password', secret + '/' + 'selfoss_db_password', length=32) }}"
 selfoss_db_database: selfoss
-selfoss_version: 2.16
+selfoss_version: 2.17
 
 selfoss_username: "{{ main_user_name }}"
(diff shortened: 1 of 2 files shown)


neuhaus:
Does it require a recent version of composer?

On Ubuntu 16.04, there is a "composer" package with version 1.0.0~beta2-1.

On Debian Stretch there is a "composer" package with version 1.2.2-1

There doesn't seem to be a "composer" package on Debian 8, could you change `roles/news/tasks/selfoss.yml` to only download composer on that distribution? 

Also I found a [guide](https://www.digitalocean.com/community/tutorials/how-to-install-and-use-composer-on-debian-8) that adds a checksum check after downloading before installing like this:

```
php -r "if (hash_file('SHA384', '/tmp/composer-setup.php') === 'sha_384_string') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('/tmp/composer-setup.php'); } echo PHP_EOL;"
```

That seems like a good idea.

PR #729, opened 2017-12-12
Install dovecot-antispam module

Titan-C:
Every time, after successfully by imap the connection would terminate with EOF and I would not get my email.

Reviewing the dovecot log there was this hint.
```
dovecot: imap: Fatal: Plugin 'antispam' not found from directory /usr/lib/dovecot/modules
```
After installing this module everything worked


mikeashley:
Thanks!

PR #738, opened 2018-04-04
Fix: Ansible uses the value present in apt module state parameter

Titan-C:
The installed value is being deprecated. This fixes that issue


neuhaus:
When was the "present" state added to ansible?

PR #740, opened 2018-04-10
Updating fail2ban jail list

bwhaines:
This changes the list of jails used by fail2ban to match the names of the config files in /etc/fail2ban/filter.d


neuhaus:
Does this work with older versions as well?

bwhaines:
This works with 0.9 and 0.10.  If we remove a few of the Apache files it'll cover 0.8 as well.

PR #741, opened 2018-04-22
Fix nested iteration of `openvpn_clients`.

johngian:


johngian:
This fixes an issue triggered after my last deployment with ``ansible 2.5.1``.

The paths when running this [1] task were wrong. Instead of iterating over the ``openvpn_clients`` values, it replaced the path with the string ``openvpn_clients`` which didn't exist.

[1] https://github.com/sovereign/sovereign/blob/master/roles/vpn/tasks/openvpn.yml#L169

PR #742, opened 2018-04-22
Disable deprecated SSLv2 in dovecot config

johngian:


johngian:
This fixes an incompatibility issue triggered when I updated my server to Debian stretch.

neuhaus:
Wouldn't this patch enable SSLv2 on older versions of Dovecot?

neuhaus:
We can apply this patch as soon as we officially drop Debian Jessie. Or we change it so it keeps the !SSLv2 on Jessie.

neuhaus:
I think we can do something like this in the template (untested):

```
# SSL protocols to use
{% if ansible_distribution_release in ["jessie", "trusty"] %}
ssl_protocols = !SSLv2 !SSLv3
{% else %}
ssl_protocols = !SSLv3
{% endif %}
```

PR #744, opened 2018-04-22
Force https on mail autoconfig apache vhost.

johngian:


johngian:
This is a patch from my local setup. It looks like Thunderbird and KMail work fine with HTTP to HTTPS redirection when trying to configure the accounts automatically.

Titan-C:
looks good to me

neuhaus:
From the documentation:

> Given the email address "fred@example.com", Thunderbird first checks <http://autoconfig.example.com/mail/config-v1.1.xml?emailaddress=fred@example.com> and then <http://example.com/.well-known/autoconfig/mail/config-v1.1.xml>.

We could get rid of this virtualhost and just use the /.well-known/ path instead...

PR #745, opened 2018-04-26
use mozilla ssh config on algorithms

Titan-C:
Following the security recommendations from mozilla https://infosec.mozilla.org/guidelines/openssh
We should update our ssh config.


neuhaus:
Ubuntu 14.04 only has OpenSSH 6.6, the recommendations by Mozilla require OpenSSH 6.7.
Perhaps we should use different templates for different versions.

PR #746, opened 2018-05-18
[Prosody] Deprecated mod_privacy in favor of blocklist

Titan-C:
After upgrading my VPS prosody wouldn't start. Error log complained mod_privacy being deprecated. I updated that. Also configuration file needs to belong to prosody user, to start the service. 

mod_privacy info and deprecation notice
https://modules.prosody.im/mod_privacy_lists.html

### DIFF
--- a/roles/xmpp/tasks/prosody.yml
+++ b/roles/xmpp/tasks/prosody.yml
@@ -26,7 +26,7 @@
   file: state=directory path=/decrypted/prosody owner=prosody group=prosody
 
 - name: Configure Prosody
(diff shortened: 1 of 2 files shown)


neuhaus:
What Linux distribution and version of Prosody do you use, @Titan-C ?

Titan-C:
Currently I'm testing on debian stretch 9.4 and prosody is 0.10.2-1~stretch1
But as specified by the prosody documentation this module is already available on prosody 0.9 even when I remember that on debian jessie I was already using prosody 0.10

neuhaus:
Does installing the prosody package create the group "prosody"?

Titan-C:
> Does installing the prosody package create the group "prosody"?

I seems it does. I didn't do that manually.

PR #753, opened 2019-03-19
Make sure roundcube tag is correctly defined

roelvanduijnhoven:
Tag is not correctly defined. Without this change, the following won't work:

> ansible-playbook --tags=roundcube

### DIFF
--- a/roles/webmail/tasks/main.yml
+++ b/roles/webmail/tasks/main.yml
@@ -1,1 +1,2 @@
-- include: roundcube.yml tags=roundcube
+- include: roundcube.yml 
+  tags: roundcube


adborden:
@roelvanduijnhoven thanks for looking into this. The inline syntax should be valid and work with `--tags`. What version of ansible are you using? It's possible that the `include` needs to be changed to `import` for newer versions.

roelvanduijnhoven:
@adborden I created this issue half a year ago, not sure what version I was using. And too bad that I did not mentioned this in the issue.

But pretty sure it was newer than the 2.2.2.0 minimum version that is mentioned in requirements.txt of this repo, as 2.3 was released in 2017-05-09.

As I am currently no longer using this repo I could not easily test this. 

However doing a quick searched brought me [to a similar topic](https://stackoverflow.com/questions/51131062/include-tags-dont-work-in-ansible). Which indeed seems to suggest that the way it is used here does _not_ work. As `tags` is parsed as an option to the module, and not the task!

Anyway: feel free to close or merge. I _think_ merging is correct.

neuhaus:
Thanks!

PR #755, opened 2019-03-25
Make tests work

roelvanduijnhoven:
I am trying to extend this repo with some functionality. The tests failed however to work. Hence this contribution that fixes the tests.

## Problem 1: bundled self-signed certificate has expired

Thus I created a new one using:

```
openssl genrsa 2048 > roles/common/files/wildcard_private.key
openssl req -new -x509 -nodes -sha1 -days 3650 -key roles/common/files/wildcard_private.key > roles/common/files/wildcard_public_cert.crt
openssl x509 -noout -fingerprint -text < roles/common/files/wildcard_public_cert.crt
cat roles/common/files/wildcard_public_cert.crt roles/common/files/wildcard_private.key > roles/common/files/wildcard_ca.pem
```

Although Python _does_ complain that this will no longer work in the future. Ideally we need to use SAN, but let's not do that now (it was also not used in the previous certificate). Error message:

> SubjectAltNameWarning: Certificate for autoconfig.sovereign.local has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/shazow/urllib3/issues/497 for details.)

## Problem 2: port 465 does not connect

Now tests that use port 465 refuse connection. However this is also the case on a truly provisioned server. So I think this is a deliberate choice, and tests were simply not updated to reflect this.

So I replace all occurrences of 465 by 587. But as this port does not explicitly start TTL, we need to do it ourselves by calling STARTTLS. 

With those two fixes the tests pass again :heavy_check_mark:.

------

Would it be an idea to create another PR that adds the instructions to README.md on how to actually run the tests?

Would love some feedback!


neuhaus:
Looks good, thanks!

PR #760, opened 2019-06-27
Use correct certificate file in prosody

lfuelling:
Prosody uses the `cert.pem` file which fails to verify on servers configured to check the certificate chain. Using the `fullchain.pem` file fixes the problem.


adborden:
Ping, would be nice to get this merged.

neuhaus:
Thanks and sorry for the delay

Issue #765, opened 2019-08-04
OwnCloud no longer supports Debiain jessie


adborden:
```
TASK [owncloud : Ensure repository key for ownCloud is in place] ***************
 [WARNING]: sftp transfer mechanism failed on [127.0.0.1]. Use ANSIBLE_DEBUG=1
to see detailed information

fatal: [jessie]: FAILED! => {"changed": false, "failed": true, "msg": "Failed to download key at https://download.owncloud.org/download/repositories/stable/Debian_8.0/Release.key: HTTP Error 404: Not Found"}
```

Issue #762, opened 2019-07-16
Vagrant up on master fails at "adding box-cutter/debian8"


chepec:
I was trying to setup the [development environment](https://github.com/sovereign/sovereign/wiki/Development-Environment).

```
$ git clone https://github.com/sovereign/sovereign.git
$ cd sovereign/
$ vagrant up
Bringing machine 'jessie' up with 'virtualbox' provider...
==> jessie: Box 'box-cutter/debian8' could not be found. Attempting to find and install...
    jessie: Box Provider: virtualbox
    jessie: Box Version: >= 0
==> jessie: Loading metadata for box 'box-cutter/debian8'
    jessie: URL: https://vagrantcloud.com/box-cutter/debian8
==> jessie: Adding box 'box-cutter/debian8' (v2.0.8) for provider: virtualbox
    jessie: Downloading: https://vagrantcloud.com/box-cutter/boxes/debian8/versions/2.0.8/providers/virtualbox.box
    jessie: Download redirected to host: cdn.boxcutter.io
An error occurred while downloading the remote file. The error
message, if any, is reproduced below. Please fix this error and try
again.

The requested URL returned error: 404 Not Found
```

On Ubuntu 18.04.2, with VirtualBox 5.2.18 (latest version from the [VirtualBox APT repo](http://download.virtualbox.org/virtualbox/debian)), Vagrant 2.2.5 ([latest deb file](https://releases.hashicorp.com/vagrant/2.2.5/vagrant_2.2.5_x86_64.deb) from their Downloads page), and Ansible v2.8.2.

I tested the URL that appears to fail, and it seems the remote file is no longer where we expect it to be? 
```
$ curl -L https://vagrantcloud.com/box-cutter/boxes/debian8/versions/2.0.8/providers/virtualbox.box
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist</Message><BucketName>smartystreets-boxcutter</BucketName><RequestId>B6F0EC33E27DF5D1</RequestId><HostId>bX/gIFt4nFtEHllNcM3/3dtpPAoIy/3BUZOei7gogq5m7oCExGruUPUMI+SvNV4RLfyKWXBsq9k=</HostId></Error
```

adborden:
https://github.com/boxcutter/debian#current-boxes
> We no longer provide pre-built binaries for these templates.

[...]

Issue #748, opened 2018-06-01
Google Auth and Prosody role ( permission denied )


gmkrab:
Please check role roles/common/tasks/google_auth.yml and roles/xmpp/tasks/prosody.yml. 

Need comment:
#become: yes
# become_user: "{{ main_user_name }}"
#when: ansible_ssh_user != "vagrant"

lfuelling:
In my case, the `Configure Prosody` task failed. The solution was to change the config file owner from `root` to `prosody`.

eg.:
```
- name: Configure Prosody
  template: src=prosody.cfg.lua.j2 dest=/etc/prosody/prosody.cfg.lua group=prosody owner=prosody
  notify: restart prosody
```

Titan-C:
Yes, @lfuelling that's one of the changes I propose on #746

PR #766, opened 2019-08-04
Fix Vagrant environment

adborden:
Fixes https://github.com/sovereign/sovereign/issues/765
Fixes https://github.com/sovereign/sovereign/issues/762

This updates the Vagrant boxes to use official Ubuntu/Debian boxes. I added a shell provisioner to add ansible dependencies (python and python-apt) which are not included in the boxes.

I ran into an error installing OwnCloud on jessie in that the deb repo is failing with a 404. I disabled the role for jessie because of that (#765). I left the owncloud test in tests.py because the test is still valid for other releases/distros, but it would be nice to move to a framework that would allow us more control over the tests running on different platforms, e.g. the test could be skipped only on jessie.

I also ran into an issue with the prosody notifier. Prosody fails to restart because it cannot read prosody.cfg.lua (permission denied) which is mentioned (https://github.com/sovereign/sovereign/issues/748#issuecomment-467603918) but there's no explicit issue for it.

### DIFF
--- a/roles/owncloud/tasks/main.yml
+++ b/roles/owncloud/tasks/main.yml
@@ -1,4 +1,9 @@
 ---
+- debug:
+    msg: >-
+      OwnCloud is no longer supported on Debian jessie
+      https://github.com/sovereign/sovereign/issues/765
+  when: ansible_distribution_release == "jessie"
 
 - include: owncloud.yml tags=owncloud
   # https://github.com/sovereign/sovereign/issues/765
--- a/roles/owncloud/tasks/main.yml
+++ b/roles/owncloud/tasks/main.yml
@@ -1,3 +1,5 @@
 ---
 
 - include: owncloud.yml tags=owncloud
+  # https://github.com/sovereign/sovereign/issues/765
+  when: ansible_distribution_release != "jessie"
--- a/roles/xmpp/tasks/prosody.yml
+++ b/roles/xmpp/tasks/prosody.yml
@@ -26,7 +26,7 @@
   file: state=directory path=/decrypted/prosody owner=prosody group=prosody
 
 - name: Configure Prosody
-  template: src=prosody.cfg.lua.j2 dest=/etc/prosody/prosody.cfg.lua group=root owner=root
+  template: src=prosody.cfg.lua.j2 dest=/etc/prosody/prosody.cfg.lua group=prosody owner=prosody
   notify: restart prosody
 
 - name: Create Prosody accounts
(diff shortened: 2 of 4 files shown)


neuhaus:
Thanks a lot!

roles/owncloud/tasks/main.yml
```diff
@@ -1,3 +1,5 @@
 ---
 
-- include: owncloud.yml tags=owncloud
\ No newline at end of file
+- include: owncloud.yml tags=owncloud
+  # https://github.com/sovereign/sovereign/issues/765
+  when: ansible_distribution_release != "jessie"
```
neuhaus: What do you think about adding a „debug“ statement to let the user know what is going on? Something like „owncloud no longer supports Debian jessie“?

roles/xmpp/tasks/prosody.yml
```diff
@@ -26,7 +26,7 @@
   file: state=directory path=/decrypted/prosody owner=prosody group=prosody
 
 - name: Configure Prosody
-  template: src=prosody.cfg.lua.j2 dest=/etc/prosody/prosody.cfg.lua group=root owner=root
+  template: src=prosody.cfg.lua.j2 dest=/etc/prosody/prosody.cfg.lua group=prosody owner=prosody
```
neuhaus: A better solution is to make the file readable by prosody (i.e. world readable) but keep root ownership.
If prosody owns its config file it can write to it (which is something it should‘nt be allowed to do)

roles/owncloud/tasks/main.yml
adborden: Sure thing.

roles/xmpp/tasks/prosody.yml
adborden: I'll test it again, but I think prosody still had issues with letting root own the file. Maybe prosody needs write access? Very strange.

You're suggesting `owner=root group=prosody mode=0664`  or something different?

adborden: I was wrong, making it world readable works, although I don't totally understand why.

adborden:
@neuhaus this is ready for another look :)

neuhaus:
Great. One more thing - is owncloud still supporting Ubuntu 14.04? If not we should show a similar message for Ubuntu trusty.
(yes, we should officially drop Ubuntu trusty now that it no longer receives updates).

I am on vacation right now so I won't have a lot of time to check here, sorry. Perhaps one of the other maintainers can step in.

neuhaus:
I'll merge this for now, I hope you can add the message for Ubuntu 14.04 if it is necessary.

adborden:
No, OwnCloud is only supporting Ubuntu 16.04, 18.04, and 19.04 at the moment.

I didn't even realize sovereign was supporting Ubuntu 14.04.

PR #769, opened 2019-08-09
Proposal: using Molecule for automated unit tests

adborden:
- [ ] This depends on https://github.com/sovereign/sovereign/pull/766, which should be merged first.

I'm using [molecule](https://molecule.readthedocs.io/en/stable/index.html) for other Ansible projects. It's very effective for development and testing of ansible roles, especially when supporting multiple platforms (jessie, stretch, xenial, etc). Because molecule can use Docker as the underlying driver, we can even run these automated tests as part of CI. This should speed up the time to review and test PRs and reduce the risk of breaking a feature or platform.

I only setup the `common` role as a proof of concept. If this approach is interesting to folks, we can take it further.

Some more features this provies:
- Automated tests using [testinfra](https://testinfra.readthedocs.io/en/latest/index.html) ([example](https://github.com/adborden/sovereign/blob/ca180fc6a0d47f7797e86c091fd8353a92acc0e5/roles/common/molecule/default/tests/test_default.py))
- Support for multiple platforms ([example](https://github.com/adborden/sovereign/blob/ca180fc6a0d47f7797e86c091fd8353a92acc0e5/roles/common/molecule/default/molecule.yml#L9))
- Molecule tests at the role level to encourage clear separation of responsibilities and reusable roles
- Multiple scenarios which allows to test with specific input variables or configurations
- Built-in linting checks
- Idempotency test (a role applied after the first time should result in no changes)

Obviously there are limitations in Docker that we cannot fully test everything, we'll still need to validate against a live server for some changes. However, this provides a pretty good baseline.

Molecule also supports Vagrant as a driver but it is considered alpha quality. This would give us better test coverage, although we might not be able to run it in CI. I'm not sure of if Travis can run VMs as part of a build.


adborden:
Here's the CI build. You'll notice that xenial fails half way. https://travis-ci.org/sovereign/sovereign/builds/570018138

adborden:
:wave: would love some input on this, if it's worth continuing. Having better automated testing would really simplify support for multiple distributions.

adborden:
Ping, trying to add [stretch](https://github.com/sovereign/sovereign/pull/778)/wheezy support involves a lot of manual testing, especially because we're trying to support so many distros in parallel. Would love any thoughts or opinions on this.

adborden:
:wave: Just looking for any feedback on this. Finding it frustrating to see so many unmerged PRs and having to do so much manual testing. Having more comprehensive automated tests would make it easier to get these open PRs merged.

lessless:
@neuhaus  this seems to be a nice idea

neuhaus:
Hi, sorry for the lack of feedback. Looks like a great idea!
We also need to limit the supported platforms since we have too few contributors.

adborden:
Quoting Sven Neuhaus (2020-03-16 08:13:51)
> Hi, sorry for the lack of feedback. Looks like a great idea!
> We also need to limit the supported platforms since we have too few contributors.

Glad to hear! I will clean this PR up. Instead of adding tests for all
roles at once, I'm going to submit a PR for each role so we can
incrementally get benefit from having _some_ tests.

--
Aaron D Borden
Human and hacker

PR #771, opened 2019-08-10
Bump ansible to 2.5

### LINKED ISSUE #761: vulnerable version of Ansible
Looks like there is a [security issue](https://nvd.nist.gov/vuln/detail/CVE-2019-3828) with Ansible <2.5.15.

We should upgrade the requirements.txt after testing that it works with an up-to-date version of Ansible. 

Any takers?

adborden:
Fixes #761

Tested against jessie in the Vagrant environment.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-ansible==2.2.2.0
+ansible~=2.5.15
 passlib


adborden:
Oops, just realized 2.5 is already end of life, and there are a number of other changes that need to happen. I'll re-submit.

PR #772, opened 2019-08-10
Fix ZNC password configuration

adborden:
- Use sha256 hash function (ZNC does not support sha512)
- Add tests for the filter plugins

Fixes #490 

The ZNC test in test.py still doesn't work. It gets through the initial assertions, but fails when initiating the IRC connection. Should I comment out the assertions that aren't working?


chrisbanm:
So I've run into the same problem and found out that now passlib version 1.7.2 is what is downloaded and the call is very different from the 1.6.x. See https://passlib.readthedocs.io/en/stable/lib/passlib.hash.sha256_crypt.html for details.  The fix specified in e4ede42 is incorrect.

krmboya:
@adborden's solution seems to work for me with passlib 1.7.2

PR #797, opened 2020-02-11
Update postgres.yml

charliemaiors:
Using state present instead of installed for package installation


neuhaus:
Thanks, this change was included in a previous pull request which got merged.

Issue #809, opened 2020-03-26
failing tests with jessie


qedi-r:
After playing with sovereign around and deploying jessie with vagrant to run the tests locally, I noticed the web tests were failing. This seemed to be that the listener monit requires should only be listening on localhost, to allow the blog VirtualHost directive to be effective. After updating `roles/monitoring/files/etc_apache2_sites-available_00-status.conf`
to 
`<VirtualHost 127.0.0.1:80>`

and updating `roles/common/tasks/apache.yml` to properly disable the default conf file, the test passed.

krmboya:
This fix has helped in the branch I'm working on in setting up for stretch.

PR #810, opened 2020-03-26
Bugfixes for issue #809

qedi-r:
Documentation for DirectoryIndex and Options on apache2 claim to be valid in VirtualHosts, but this never actually took effect for me until I put it in the Directory block. Clarifying the listener for monit was also required, and should be safer for future configuration changes.

In jessie, apache uses `.conf`` name endings.


qedi-r:
Should address #809

Issue #789, opened 2020-01-25
Add support for Ubuntu 'Focal Fossa' 20.04 LTS


neuhaus:
It will be released in April 2020.

I have updated the [wiki pages with the versions](https://github.com/sovereign/sovereign/wiki/Software-used-by-Sovereign).

neuffer:
Are there any issues known currently?

neuhaus:
Issues affecting buster will also affect focal and there may be some additional ones

PR #813, opened 2020-05-15
Ubuntu 20.04 (focal) support

tbaumann:
Mainly python3 pains.

Using Ubuntu rspamd because there is no upstream repo.

This partly fixes https://github.com/sovereign/sovereign/issues/789

Untested modules, those will likely still fail
    - news
    - git
    - readlater
    - blog
    - ircbouncer
    - xmpp
    - owncloud
    - vpn

### DIFF
--- a/roles/common/tasks/google_auth.yml
+++ b/roles/common/tasks/google_auth.yml
@@ -5,8 +5,6 @@
   apt: pkg={{ item }} state=present
   with_items:
     - libpam-google-authenticator
-    - libpam0g-dev
-    - libqrencode3
   tags:
     - dependencies
 
--- a/roles/common/tasks/letsencrypt.yml
+++ b/roles/common/tasks/letsencrypt.yml
@@ -19,8 +19,29 @@
     owner=root
     group=root
 
+- name: Install letsencrypt python dependencies
+  when: ansible_facts['distribution_release'] == 'focal'
+  apt: pkg={{ item }} state=present
+  with_items:
+    - python-is-python3
+    - python3-venv
+    - python3-virtualenv
+    - python3-dev
+    - libssl-dev
+    - gcc
+    - openssl
+    - libffi-dev
+    - ca-certificates
+
 - name: Install LetsEncrypt package dependencies
+  when: ansible_facts['distribution_release'] != 'focal'
   command: /root/letsencrypt/letsencrypt-auto --help
+  register: le_deps_result
+  changed_when: "'Bootstrapping dependencies' in le_deps_result.stdout"
+
+- name: Install LetsEncrypt package dependencies
+  when: ansible_facts['distribution_release'] == 'focal'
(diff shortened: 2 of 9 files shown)


tbaumann:
I'm not sure I will continue to invest into this project. Ansible isn't my strength and I find this all quite rough around the edges. I think I go back to setting it up manually the way I like it.

adborden:
Related: [letsencrypt/cerbot is deprecating certbot-auto](https://github.com/certbot/certbot/issues/7941), so the manual installtion of depedencies here will be necessary on other distributions as well.

adborden:
@tbaumann thank you for your efforts! I understand what you mean about the rough edges. As we've added support for newer distributions and features, I feel like there's been regressions in other places. Hopefully we can improve that soon.

I'm testing this out and I've hit a few snags. It could just be differences in our base images. I'm using Ubuntu 20.04 on Google Cloud. What were you testing on?

tbaumann:
I used the cloud image from my vps provider.
I don't think it's significantly modified. And I ran Dist-Upgrade before

tbaumann:
I think the issues I had to deal with are basically one of two categories.

Either different package sources or python3 snafu.

tbaumann:
It's been a while. Rough around the edges was mostly that I wasn't confident that I pulled out all the hard coded credentials.
And the way users were added was beyond confusing to me.
Somehow they exist as passwd users but also in this SQL database. I never managed to log into the web portal.
And the way the SQL uses are inserted didn't inspire me with confidence that it's safe to run this task multiple times to add users.

Basically I decided that there is too much hidden complexity and that I'm better off doing it the traditional way.
But admittedly that task has been on my smelly pile of things to do ever since. So maybe I should have pushed through.

adborden:
Ah, I see what you mean about the rough edges. Sovereign is pretty opinionated about how it does things, so you're giving up some flexibility in exchange for automated configuraiton.

In general, the playbook is idempotent*, so running it multiple times is fine, even encouraged because the idea is that running the playbook will get your instance into a known good state (at least in theory).

Re: users, sovereign creates one main OS user with SSH access. For mail, IRC, prosody, etc, sovereign creates virtual users. Out of the box, Postfix uses local users on the system, but sovereign configures it to use [virtual users and mail domains from PostgreSQL](http://www.postfix.org/VIRTUAL_README.html).

_* It's been a while but the wallbag/readlater role may not be idempotent, but running it multiple times will fail harmlessly._

adborden:
I got stuck on the `webserver` role. PHP PEAR is not installed. Even after installing it, the `pear` task fails. This is a module in early preview state, so it's not really considered stable. I think we'll want to fall back to the `command` module, instead.

Is PEAR the only way to get php-mcrypt on focal?

tbaumann:
Pear just seemed the easiest way. I guess the alternative would be to build the module locally.
Don't know though.

Issue #815, opened 2020-05-24
Dovecot can't create maildir subfolders because of encfs


lfuelling:
HI, 
this is more of an FYI issue because I already fixed it for me and I doubt that there is an easy general fix for this.

I tried to create mailbox subfolders in roundcube (and later using `doveadm`) but always got an error like the following:
```
May 24 10:38:06 host dovecot: imap(mail@host.tld)<16086><oFwsdhKmWLqU+xSn>: Error: Couldn't create mailbox list lock /decrypted/host.tld/mail/mailboxes.lock: file_create_locked(/decrypted/host.tld/mail/mailboxes.lock) failed: link(/decrypted/host.tld/mail/mailboxes.lock22f7b6bc4aa261b8, /decrypted/host.tld/mail/mailboxes.lock) failed: Operation not permitted
```

After hours of searching the web and trying out different permissions for the maildir, I was able to fix this issue by ditching encfs completely. 
I basically just copied the `/decrypted` folder somewhere else, deleted the mount and the encrypted directory, moved everything back into place and deleted the encfs steps from the ansible scripts.

I was able to do this because the server I use has disk encryption enabled. And this is also why I don't think that there is an easy fix for this that works for everybody.

But if you have the same error and ditching encfs is okay for your setup, this might be the fix you need.

func0der:
Hey @lfuelling ,

this might be the wrong place for this issue.
Please open an issue over at https://github.com/dovecot/core/, so everyone that searches in the right place can find this issue (if one).

Thank you.

PR #817, opened 2020-05-25
Fix  active sieve showing up as mailbox dir

lfuelling:
After fixing #815, the active sieve is showing up as a mailbox folder in roundcube.

This is happening because the mailboxes are stored like `${maildir}/.INBOX.foldername` and the active sieve is located at `${maildir}/.dovecot.sieve`. So there will be a `dovecot/sieve` folder showing up in IMAP clients.

This PR fixes that by moving the active sieve to `${maildir}/sieve/.dovecot.sieve`.

I have tested this for two days now and it's working like a charm on my setup. The active sieve file also doesn't show up as a filter group in roundcube.


neuhaus:
Looks good. What happens with existing installations?

lfuelling:
I'm nut sure but I was not able to see any bugs or glitches on my setup (I did this on an exisiting installation).

PR #819, opened 2020-07-26
Tuning cgit config

Titan-C:
I was performing a new installation on ubuntu focal and realized at least for the cgit part, this changes provided a good configuration.

Ensure idempotent git role
- Add cgit dependency which was missing
- Move pip install to common
- Update versions


neuhaus:
Looks good! I don't use gitolite myself so hopefully someone else will have a look. If not I'll merge it in a week or two.

PR #778, opened 2019-09-28
Stretch support

adborden:
Resolves #695. This is a work in progress that adds Debian Stretch support. These changes add stretch support without removing support for jessie.

Remaining tasks:
- [ ] Ansible 2.6+ (PR #773)
- [ ] ZNC service will not start (PR #772)
- [ ] Cannot log into roundcube (#774)
- [ ] Missing PHP dependencies for OwnCloud (#775)

PR #820, opened 2020-08-13
Adds Support for Debian Stretch

programmer-ke:
# What
Add support for Debian 9 (Stretch)

# Why
Debian 8 (Jessie), the currently supported version has been EOL'd

https://www.debian.org/News/2020/20200709

# How

- Starting point was @adborden branch in this PR: https://github.com/sovereign/sovereign/pull/778
- On top of that, I applied the commits that had since gone into master.
- Applied Apache config fixes by @qedi-r in an open PR: https://github.com/sovereign/sovereign/pull/810
- Applied @adborden ZNC fix: https://github.com/sovereign/sovereign/commit/6cae4fe60a3676190ba01a3bcf1d45b6cb87871f
- Additional fixes to get the playbook run to completion in Stretch.

# What works
I'll be updating this as I test further

- [x] Tarsnap Backups
- [x] Webhosting
- [x] Roundcube webmail (However, email testing with verifier.port25.com returns None for the DKIM check)
- [x] Monit
- [x] Git hosting 
- [x] Selfoss Feeds
- [x] ZNC
- [ ] VPN
- [x] Owncloud
- [ ] Wallabag
- [ ] XMPP

Feedback and suggestions very much welcomed


kennyrc:
A complication with getting owncloud working is that the latest version supports php 7.2 and above, while the default php installation is 7.0. 

Currently trying to see if updating the default php version used by Apache will, help (and hopefully play well with the other php webapps).

kennyrc:
Detected an issue with znc password generation: https://github.com/kennyrc/sovereign/blob/stretch-dev-kr/roles/ircbouncer/templates/usr_lib_znc_configs_znc.conf.j2#L70

`passlib.hash.sha256_crypt.encrypt` is called twice, each time returning different values for hash and salt. To fix

programmer-ke:
Long time coming,but finally managed to solve the znc password generation.

The first issue with this was a bug in populating the password and salt, because the hash function was called twice, once each time for salt and hash, this meant the salt and hash never matched.

Even after fixing the above, discovered that the values produced did not work with znc's internal hash function. I could not figure out the correct `rounds` parameter to pass to `passlib.hash.sha256_crypt.encrypt` to make it work.

The solution was to therefore generate these values interactively using znc itself. I used ansible's built in `expect` module for this. Inspiration was from: https://git.xythobuz.de/thomas/sovereign/commit/484ee1eb994dc45e84c37665767d3082c953da52

programmer-ke:
An issue I've identified with wallabag is that login does not work, even with the correct credentials. I was able to use the same credentials from `psql` and they worked, so not sure why it doesn't work via the web interface. 

However, I've noticed we are installing an outdated version of wallabag, from back in 2015. Will update it to the latest version and see if it'll work.

programmer-ke:
Possible reason for wallabag authentication not working is because of the missing deprecated `mcrypt` extension in php7.2. Moving to php 7.2 was necessary since this is the minimum supported version by owncloud.

There are workarounds for installing `mcrypt` in php 7.2 but since it's discouraged to continue depending on the deprecated package, it's ultimately better to install a later version of wallabag without the dependency.

The best course of action seems to be to install the latest version of wallabag for those who don't already have it. For those who already do and are upgrading, it may be best to export wallabag v1 data and import it again after re-installing the latest version.

neuhaus:
Thanks for your efforts. If we run into a hard problem with individual packages I'm open to dropping some of them due to a lack of manpower. Perhaps it makes sense to skip Debian Stretch and work on Debian Buster instead?

programmer-ke:
> I'm open to dropping some of them due to a lack of manpower. 

Thanks, I'll keep that in mind.

> Perhaps it makes sense to skip Debian Stretch and work on Debian Buster instead?

Yeah, that would be ideal. I initially tried Buster but too many things were breaking. I figured if we could get it working with Stretch, then making to leap to Buster would be easier.

neuhaus:
We could also aim for Debian 11 "bullseye" instead, it will be released soon (issue #824 )
Instead of having several large pull requests i think it's have better break them up into multiple PRs (usually one per service).
I'm also OK with dropping support for previous distributions and Ubuntu due to  lack of manpower.

programmer-ke:
Sorry folks, ran out of steam with this, after I ran into some PHP dependency issues that were tricky to resolve. I decided I just wanted something simpler that relied purely on Debian's default repositories without this dependency hell.

Worked on a smaller system from scratch just for mail and websites (https://github.com/programmer-ke/replatform/)

I'll  just close this.

PR #823, opened 2021-03-17
Simplify useradd instructions in README

benpbolton:
`-m` or `--create-home` flags in useradd will automatically create a home folder with user appropriate permissions.


neuhaus (approved):
Thanks!

PR #833, opened 2022-12-11
configuring two factor

tirth78:
Tried to fix the issue of not storing the keys even after having valid ssh.


neuhaus:
Hi, why do you think /tmp/ is not a good location? Because it is gone after a reboot? Then we could move it to /var/tmp/

The user should remove those files anyway.
`/keys_stored` does not exist on a normal system.

